PR #1, opened 2020-04-28
IO4: Observer for a driver changes events

asmei1:


src/main/java/edu/kis/powp/jobs2d/drivers/DriverChangeObserver.java
```diff
@@ -0,0 +1,11 @@
+package edu.kis.powp.jobs2d.drivers;
+
+import edu.kis.powp.jobs2d.features.DriverFeature;
+import edu.kis.powp.observer.Subscriber;
+
+public class DriverChangeObserver implements Subscriber {
```
kowallus: The name should be more about, what it does (not only what it observers).

kowallus (changes requested):
Really nice. Only the observer name should be more specific.

src/main/java/edu/kis/powp/jobs2d/drivers/DriverChangeObserver.java
asmei1: Will DriverChangeTitleObserver be better name?

PR #2, opened 2020-04-28
Visitor pattern in command classes hierarchy

3elk4:


src/main/java/edu/kis/powp/jobs2d/command/DriverCommandVisitor.java
```diff
@@ -0,0 +1,31 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.Iterator;
+
+
+/**
+ * Visitor class for driver commands.
+ */
+public class DriverCommandVisitor implements Visitor {
```
kowallus: The name should indicate also the task of the visitor.

src/main/java/edu/kis/powp/jobs2d/command/DriverCommandVisitor.java
```diff
@@ -0,0 +1,31 @@
...
+    private int counter = 0;
+
+    public int getCounter(){
+        return this.counter;
+    }
+
+    @Override
+    public void visit(DriverCommand driverCommand) {
+        counter++;
+        System.out.println("Counter: " + counter);
```
kowallus: Please remove such side effects. Printing the result obtained from getCounter will be enough.

src/main/java/edu/kis/powp/jobs2d/command/DriverCommandVisitor.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public void visit(ICompoundCommand compoundCommand) {
+        Iterator<DriverCommand> iterator = compoundCommand.iterator();
```
kowallus: This solution will not be correct if compound command contains another compound command. 
There is a generic way to solve it using visitor in the context of composite.

src/main/java/edu/kis/powp/jobs2d/command/DriverCommandVisitor.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Iterator;
+
+
+/**
+ * Visitor class for driver commands.
+ */
+public class DriverCommandVisitor implements Visitor {
+    private int counter = 0;
+
+    public int getCounter(){
```
kowallus: our convention is to use tabs instead of spaces

src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
```diff
@@ -6,7 +6,6 @@
  * Interface extending Job2dDriverCommand to execute more than one command.
  */
 public interface ICompoundCommand extends DriverCommand {
```
kowallus: It should be visitable!

src/main/java/edu/kis/powp/jobs2d/command/Visitable.java
```diff
@@ -0,0 +1,8 @@
+package edu.kis.powp.jobs2d.command;
+
+/**
+ * Interface for objects using visitor.
+ */
+public interface Visitable {
```
kowallus: It should not be a separate interface because Visitor class applies only to commands hierarchy. Please move it to DriverCommand.

src/test/java/edu/kis/powp/jobs2d/TestCompoudCommand.java
```diff
@@ -0,0 +1,29 @@
...
+import edu.kis.powp.jobs2d.command.*;
+
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Test class for ICompoundCommand interface.
+ */
+public class TestCompoudCommand implements ICompoundCommand, Visitable {
```
kowallus: I do not see the reason for tests to implement these interfaces

src/test/java/edu/kis/powp/jobs2d/TestCompoudCommand.java
```diff
@@ -0,0 +1,29 @@
...
+    private List<DriverCommand> commands = Arrays.asList(new SetPositionCommand(0,0), new SetPositionCommand(0,0), new SetPositionCommand(0,0));
```
kowallus: I do not understand the construction of this test.

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -47,6 +53,16 @@ private static void setupCommandTests(Application application) {
 
 		application.addTest("Run command", new SelectRunCurrentCommandOptionListener(DriverFeature.getDriverManager()));
 
+		application.addTest("Visitor test DriverCommand", event -> {
+			DriverCommand command = new OperateToCommand(0, 0);
+			Visitor visitor = new DriverCommandVisitor();
+			((OperateToCommand) command).accept(visitor);
```
kowallus: With visitor pattern casting shouldn't be necessary. Fixing another issues may help.

kowallus (changes requested):
The solution requires several fixes. They should simplify the code.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -42,6 +43,11 @@ public void execute(Job2dDriver driver) {
 				driverCommands.forEach((c) -> c.execute(driver));
 			}
 
+			@Override
+			public void accept(Visitor visitor) {
+
```
kowallus: please add default accept implementation

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -55,13 +54,32 @@ private static void setupCommandTests(Application application) {
...
+				}
+
+				@Override
+				public void accept(Visitor visitor) {
+					commands.forEach(command -> command.accept(visitor));
+				}
+			};
+			DriverCommandCallCounterVisitor visitor = new DriverCommandCallCounterVisitor();
 			command.accept(visitor);
+			System.out.println("Counter: " + visitor.getCounter());
```
kowallus: The modification in TestJobs2dApp is to large, move it to separate listener class.

Print should use application gui, e.g. logger or window popup.

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -55,13 +54,32 @@ private static void setupCommandTests(Application application) {
...
+				}
+
+				@Override
+				public void execute(Job2dDriver driver) {
+
+				}
+
+				@Override
+				public void accept(Visitor visitor) {
+					commands.forEach(command -> command.accept(visitor));
```
kowallus: This is no a correct accept implementation. You can provide the interface ICompoundCommand with default implementation of accept. 
Iteration should be a part of visit(ICompoundCommand) method in concrete visitors.

src/main/java/edu/kis/powp/jobs2d/command/DriverCommandCallCounterVisitor.java
```diff
@@ -0,0 +1,20 @@
+package edu.kis.powp.jobs2d.command;
+
+import java.util.Iterator;
+
+
+/**
+ * Visitor class for driver commands.
+ */
+public class DriverCommandCallCounterVisitor implements Visitor {
```
kowallus: Please rename Visitor to DriverCommandVisitor.

src/main/java/edu/kis/powp/jobs2d/command/DriverCommandCallCounterVisitor.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public class DriverCommandCallCounterVisitor implements Visitor {
+	private int counter = 0;
+
+	public int getCounter(){
+		return this.counter;
+	}
+
+	@Override
+	public void visit(DriverCommand driverCommand) {
```
kowallus: IMPORTANT: Visit method should not be declared for the top of the hierarchy, but for particular commands and compound command. 
This is crucial for implementing visitors. The single call to visit (using accept) should trigger calculation of correct counter value.

kowallus (changes requested):
Your visitor is still built incorrectly. See remarks.

src/main/java/edu/kis/powp/jobs2d/command/DriverCommandCallCounterVisitor.java
```diff
@@ -1,20 +1,37 @@
...
 		return this.counter;
 	}
 
 	@Override
-	public void visit(DriverCommand driverCommand) {
-	    counter++;
+	public void visit(ICompoundCommand driverCommand) {
+		Iterator<DriverCommand> commandsIterator = driverCommand.iterator();
+		while (commandsIterator.hasNext()) {
+			DriverCommand partOfCompoundCommand = commandsIterator.next();
```
kowallus: try foreach loop, much simpler and readable:
for (DriverCommand partOfCompoudCommand: driverCommand.iterator())

src/main/java/edu/kis/powp/jobs2d/command/DriverCommandCallCounterVisitor.java
```diff
@@ -1,20 +1,37 @@
...
+			if (partOfCompoundCommand instanceof ICompoundCommand) {
+				this.visit((ICompoundCommand)partOfCompoundCommand);
+			}
```
kowallus: Visitor pattern makes casting unnecessary, but you should always use accept (never call visit directly).

src/main/java/edu/kis/powp/jobs2d/command/DriverCommandCallCounterVisitor.java
```diff
@@ -1,20 +1,37 @@
...
+			this.counter++;
```
kowallus: this is actually a form of code duplication - finally should be removed for the visitor to work correctly

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -52,35 +52,15 @@ private static void setupCommandTests(Application application) {
...
-
-				@Override
-				public void accept(Visitor visitor) {
-					commands.forEach(command -> command.accept(visitor));
-				}
-			};
-			DriverCommandCallCounterVisitor visitor = new DriverCommandCallCounterVisitor();
-			command.accept(visitor);
-			System.out.println("Counter: " + visitor.getCounter());
+			logger.info("Counter: " + visitor.getCounter());
```
kowallus: extract this lambda to separate file (too much changes in TestJobs2dApp)

kowallus (changes requested):
You are getting close to the correct solution!

kowallus (approved):
OK! Please, resolve conflicts and merge.

PR #4, opened 2020-04-29
A deep copy of a complex command

Hergoln:


src/main/java/edu/kis/powp/jobs2d/command/CommandCopier.java
```diff
@@ -0,0 +1,27 @@
...
+				Iterator<DriverCommand> iter = ((ICompoundCommand) toCopy).iterator();
+				List<DriverCommand> commandList = new ArrayList<>();
+				while (iter.hasNext()) {
+					commandList.add(iter.next().clone());
+				}
+				((ICompoundCommand) copy).setCommands(commandList);
+			}
+		} catch (CloneNotSupportedException exc) {
+			System.out.println(exc);
+		}
```
kowallus: I think that we should throw this exception outside rather return null. It might be confusing to find the reason of NullPointerException rather than CloneNotSupported.

src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
```diff
@@ -9,4 +10,5 @@
 
 	public Iterator<DriverCommand> iterator();
 
+	public void setCommands(List<DriverCommand> collect);
```
kowallus: IMPORTANT: I cannot agree to this modification. ICompoundCommand should not assume mutability!

src/main/java/edu/kis/powp/jobs2d/command/CommandCopier.java
```diff
@@ -0,0 +1,27 @@
...
+
+		DriverCommand copy = null;
+
+		try {
+			copy = toCopy.clone();
+			if (toCopy instanceof ICompoundCommand) {
+				Iterator<DriverCommand> iter = ((ICompoundCommand) toCopy).iterator();
+				List<DriverCommand> commandList = new ArrayList<>();
+				while (iter.hasNext()) {
+					commandList.add(iter.next().clone());
```
kowallus: I wonder: what would be the difference between calling clone() and copyCommand (recursively). I think that default implementation of clone can be a part of ICompoundCommand interface (using iterator operation and clone).

src/test/java/edu/kis/powp/jobs2d/events/SelectCopySecretCommand.java
```diff
@@ -0,0 +1,80 @@
...
+		commands.add(new OperateToCommand(70, 50));
+		commands.add(new OperateToCommand(20, 50));
+		compound.setCommands(commands);
+		third = CommandCopier.copyCommand(compound);
+
+		List<DriverCommand> toCopy = new ArrayList<>();
+		toCopy.add(first);
+		toCopy.add(second);
+		toCopy.add(third);
+
```
kowallus: To check deep copy we should test cloning CompoundCommand made of CompoundCommands!

src/test/java/edu/kis/powp/jobs2d/events/SelectCopySecretCommand.java
```diff
@@ -0,0 +1,80 @@
...
+		@Override
+		public DriverCommand clone() throws CloneNotSupportedException {
+			return (DriverCommand) super.clone();
+		}
+
+		@Override
+		public void setCommands(List<DriverCommand> driverCommands) {
+			this.driverCommands = driverCommands;
+		}
+	};
```
kowallus: Maybe it will be easier to create a separate implementation of ICompoundCommand in commands hierarchy.

kowallus (changes requested):
I have pointed out several issues that should be considered in the solution to the task.

src/test/java/edu/kis/powp/jobs2d/events/SelectCopySecretCommand.java
Hergoln: Maybe, but creation of implementation of ICompundCommand is a task given to another team.

kowallus: Yes, but they have to implement a specific ICompoundCommand (immutable). You can always add your own independent implementation.

src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
```diff
@@ -8,7 +9,15 @@
...
+	Iterator<DriverCommand> iterator();
 
-	public void setCommands(List<DriverCommand> collect);
+	@Override
+	default DriverCommand clone() throws CloneNotSupportedException {
+		List<DriverCommand> commands = new ArrayList<>();
+		Iterator<DriverCommand> iter = iterator();
+		while (iter.hasNext()) {
+			commands.add(iter.next().clone());
+		}
```
kowallus: If you could add Iterable interface to this interface then we could use foreach loop, more readable:
for (DriverCommand command: this)
Right?

kowallus (changes requested):
OK. Could you please try one more modification?

src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
Hergoln: Done in next commit

PR #6, opened 2020-05-05
MACRO

Lefiuu:


src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -73,7 +73,7 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
 		content.add(btnClearObservers, c);
 	}
 
-	private void clearCommand() {
+	public void clearCommand() {
```
kowallus, line 76: Why visibility is changed?!

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -66,16 +81,41 @@ public synchronized DriverCommand getCurrentCommand() {
...
+		commands.add(new OperateToCommand(i, i1));
+	}
+	public List<DriverCommand> getCommands(){
+		return commands;
+
+	}
+	public void clearCommands(){
+		commands.clear();
+		names.clear();
+	}
```
kowallus, line 119: Try extracting this part to a separate driver class and use it from the manager.

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigureOptionListener.java
```diff
@@ -17,5 +19,8 @@ public SelectTestFigureOptionListener(DriverManager driverManager) {
 	@Override
 	public void actionPerformed(ActionEvent e) {
 		FiguresJoe.figureScript1(driverManager.getCurrentDriver());
+		DriverCommandManager manager = CommandsFeature.getDriverCommandManager();
+		FiguresJoe.figureScript1(manager);
+		manager.setCurrentCommand(manager.getCommands(),"FigureJoe1");
```
kowallus, line 24: You should not modify those tests. It should be not necessary to test the feature.

kowallus (changes requested):
The code makes too many changes in existing classes! 
NOTE: Revert changes carefully.

PR #11, opened 2020-05-06
Convert mouse clicks to driver calls, after fixes second time

Trackres:


src/main/java/edu/kis/powp/jobs2d/features/MouseDrawFeature.java
```diff
@@ -0,0 +1,32 @@
...
+
+import edu.kis.powp.appbase.Application;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import javax.swing.*;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
+
+public class MouseDrawFeature {
+	private static final int WIDTH_REL = 268;
```
kowallus, line 11: RELATIVE to what?

kowallus (approved):
I am not fully satisfied, but it is merge'able :) congrats!

PR #12, opened 2020-05-06
Io 4 dev, MACRO feature

Lefiuu:


src/main/java/edu/kis/powp/jobs2d/command/manager/MacroCommandManager.java
```diff
@@ -0,0 +1,121 @@
...
+    List<List<DriverCommand>> macro= new ArrayList<>();
+    /**
+     * Set current command.
+     *
+     * @param commandList Set the command as current.
+     */
+    public synchronized void setCurrentCommand(DriverCommand commandList) {
+        this.currentCommand = commandList;
+        changePublisher.notifyObservers();
+    }
```
kowallus, line 32: This class has too many responsibilities. There is already DriverCommandManager available which we can use to set any command and execute it ("Run command"). Macro feature duplicates. 
Please reuse DriverCommandManager in this responsibility and simplify MacroCommandManager (the name should probably be changed after that).

src/test/java/edu/kis/powp/jobs2d/events/SelectTestFigure2OptionListener.java
```diff
@@ -17,5 +19,8 @@ public SelectTestFigure2OptionListener(DriverManager driverManager) {
 	@Override
 	public void actionPerformed(ActionEvent e) {
 		FiguresJoe.figureScript2(driverManager.getCurrentDriver());
+		MacroCommandManager manager = MacroFeature.getDriverCommandManager();
+		FiguresJoe.figureScript2(manager);
+		manager.setCurrentCommand(manager.getCommands(),"FigureJoe2");
```
kowallus, line 24: Figure tests should not be changed at all! MACRO feature can be transparent, because macro recorder is a driver (it can be made available through driverManager.getCurrentDriver())

kowallus (changes requested):
Better, but not enough.

PR #14, opened 2020-05-08
Created Immutable implementation of ICompoundCommand

startled-cat:


src/main/java/edu/kis/powp/jobs2d/command/ImmutableCompoundCommand.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+
+public class ImmutableCompoundCommand implements ICompoundCommand {
+
+    private List<DriverCommand> driverCommands = null;
+
+    public ImmutableCompoundCommand(List<DriverCommand> driverCommands) {
+        this.driverCommands = Collections.unmodifiableList(driverCommands);
```
kowallus, line 14: driverCommands can consist of another ICompoundCommands which may not be protected from changes. You should use "A deep copy of a complex command" feature to make sure that references to internal ICompoundCommands instances will be copied. 
Tip: it is as simple as cloning ;)

kowallus (changes requested):
Not tested and probably not 100% correct.

PR #15, opened 2020-05-10
Implementation of ImmutableCompoundCommand

startled-cat:
In comparison to previous pr, added test and deep copying in ImmutableCompoundCommand constructor.

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/ImmutableCompoundCommand.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/ImmutableCompoundCommand.java
@@ -9,26 +9,19 @@ import java.util.List;
 
 public class ImmutableCompoundCommand implements ICompoundCommand {
 
-    private List<DriverCommand> driverCommands = null;
+    private List<DriverCommand> driverCommands;
 
     public ImmutableCompoundCommand(List<DriverCommand> driverCommands) {
 
         this.driverCommands = new ArrayList<>();
-        for (Object object : driverCommands) {
-            if (object instanceof DefaultCompoundCommand) {
-                List<DriverCommand> dc = new ArrayList<>();
-                for (DriverCommand command : (DefaultCompoundCommand) object) {
-                    dc.add(command);
-                }
-                ImmutableCompoundCommand icc = new ImmutableCompoundCommand(dc);
-                this.driverCommands.add(icc);
-
-            }else{
-                this.driverCommands.add((DriverCommand) object);
+        for (DriverCommand command : driverCommands) {
+            try {
+                this.driverCommands.add(command.clone());
+            }catch (CloneNotSupportedException e){
+                this.driverCommands.add(command);
             }
         }
         this.driverCommands = Collections.unmodifiableList(this.driverCommands);
-
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/test/java/edu/kis/powp/jobs2d/command/ImmutableCompoundCommandTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Test
+    void iterator() {
+
+        List<DriverCommand> dcList = new ArrayList<>();
+        SetPositionCommand spc = new SetPositionCommand(10, 10);
+        dcList.add(spc);
+        dcList.add(new OperateToCommand(10, 10));
+        dcList.add(new SetPositionCommand(10, 10));
+        dcList.add(new OperateToCommand(10, 10));
```
kowallus, line 23: To test "deep immutability" create and add a mutable simple ComplexComand inside this CompoundCommand and then change it.

src/main/java/edu/kis/powp/jobs2d/command/ImmutableCompoundCommand.java
```diff
@@ -11,7 +12,9 @@
     private List<DriverCommand> driverCommands = null;
 
     public ImmutableCompoundCommand(List<DriverCommand> driverCommands) {
-        this.driverCommands = Collections.unmodifiableList(driverCommands);
+        this.driverCommands = new ArrayList<>();
+        this.driverCommands.addAll(driverCommands);
+        this.driverCommands = Collections.unmodifiableList(this.driverCommands);
```
kowallus: This doesn't change much. Cooperate with the group responsible for "A deep copy of a complex command" to fix the solution.

kowallus (changes requested):
The test is nice, but fixes insufficient.

src/test/java/edu/kis/powp/jobs2d/command/ImmutableCompoundCommandTest.java
startled-cat: Test updated

src/main/java/edu/kis/powp/jobs2d/command/ImmutableCompoundCommand.java
startled-cat: Corrected deep copying in latest commit.

src/main/java/edu/kis/powp/jobs2d/command/ImmutableCompoundCommand.java
```diff
@@ -12,9 +12,23 @@
     private List<DriverCommand> driverCommands = null;
 
     public ImmutableCompoundCommand(List<DriverCommand> driverCommands) {
+
         this.driverCommands = new ArrayList<>();
-        this.driverCommands.addAll(driverCommands);
+        for (Object object : driverCommands) {
+            if (object instanceof DefaultCompoundCommand) {
```
kowallus: object-oriented design makes often possible to avoid instanceof. This is the case.

kowallus (changes requested):
Have you considered "deep copy" from done by another group?

PR #16, opened 2020-05-10
Add visitor pattern to commands, with test

bopke:


src/main/java/edu/kis/powp/jobs2d/command/CommandVisitor.java
```diff
@@ -0,0 +1,27 @@
+package edu.kis.powp.jobs2d.command;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class CommandVisitor implements ICommandVisitor {
```
kowallus, line 5: The name should provide info about the function introduced with this particular visitor.
Executing a command is already in API o command so this visitor is redundant. Change this function to something new.

src/test/java/edu/kis/powp/jobs2d/TestCommandVisitor.java
```diff
@@ -0,0 +1,112 @@
...
+            @Override
+            public void accept(ICommandVisitor commandVisitor) {
+                for (DriverCommand elem : driverCommands) {
+                    if (elem instanceof SetPositionCommand)
+                        commandVisitor.visit((SetPositionCommand) elem);
+                    if (elem instanceof OperateToCommand)
+                        commandVisitor.visit((OperateToCommand) elem);
+                    if (elem instanceof ICompoundCommand)
+                        commandVisitor.visit((ICompoundCommand) elem);
+                }
```
kowallus, line 50: incorrect! Implementation of accept should always be the same (it can be put in the ICompoundCommand interface).

src/test/java/edu/kis/powp/jobs2d/TestCommandVisitor.java
```diff
@@ -0,0 +1,112 @@
...
+        public void visit(OperateToCommand operateToCommand) {
+            operateToVisits++;
+            operateToCommand.execute(job2dDriver);
+        }
+
+        @Override
+        public void visit(ICompoundCommand compoundCommand) {
+            compoundCommandVisits++;
+            compoundCommand.accept(this);
+        }
```
kowallus, line 109: Put it in a separate class (instead of CommandVisitor class?)

src/test/java/edu/kis/powp/jobs2d/TestCommandVisitor.java
```diff
@@ -0,0 +1,112 @@
...
+        @Override
+        public void visit(OperateToCommand operateToCommand) {
+            operateToVisits++;
+            operateToCommand.execute(job2dDriver);
+        }
+
+        @Override
+        public void visit(ICompoundCommand compoundCommand) {
+            compoundCommandVisits++;
+            compoundCommand.accept(this);
```
kowallus, line 108: This is not the way of using the visitor on composite objects (use iterator).

kowallus (changes requested):
Almost ok.

kowallus:
Do not move the task to "DONE" unless merged (you can use BLOCKED column)

PR #19, opened 2020-05-11
IO-1 Extending Command Manager window

Miretpl:
Patryk Janowski
Przemysław Mirowski


pom.xml
```diff
@@ -14,8 +14,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.0</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>10</source>
+                    <target>10</target>
```
kowallus, line 18: Is this really necessary?

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -42,12 +42,7 @@ private static void setupPresetTests(Application application) {
      *
      * @param application Application context.
      */
-    private static void setupCommandTests(Application application) {
-        application.addTest("Load secret command", new SelectLoadSecretCommandOptionListener());
-
-        application.addTest("Run command", new SelectRunCurrentCommandOptionListener(DriverFeature.getDriverManager()));
-
-    }
+    private static void setupCommandTests(Application application) { application.addTest("Load secret command", new SelectLoadSecretCommandOptionListener()); }
```
kowallus, line 45: Revert ALL modifications in this file.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -68,6 +69,10 @@ public synchronized void clearCurrentCommand() {
         currentCommand = null;
     }
 
+    public synchronized void runCurrentCommand() { currentCommand.execute(job2dDriver); }
```
kowallus, line 72: fix formatting (make similar to the rest of the code)

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -13,6 +13,7 @@
  */
 public class DriverCommandManager {
     private DriverCommand currentCommand = null;
+    private Job2dDriver job2dDriver = null;
```
kowallus, line 16: do not "cache" the driver. Always use driverManager to get current instance.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
kowallus, line 72: I think job2dDriver may be obsolete

kowallus (changes requested):
Please fix issues and resolve conflicts with main IO-1 branch

PR #20, opened 2020-05-12
Extending Command Manager window

k-chr:
* Moved "User Test" -> "Run Command" to Command Manager window.
* Added "Reset Observers" button.
* Implemented all needed behaviour to run these actions invoked by buttons:
   * subscribers are cached when deleted, to perform optional restore.
   * added reference to _DriverManager_ from _DriverFeature_ to _DriverCommandManager_ to get  current driver. 
   * hid (nearly) direct usages of _changePublisher_  

Q to reviewer:
- Would be fine if _changePublisher_ remained completely hidden? I mean, programmer has direct access to modifiable list of subscribers, so he/she could add/remove elements from it without direct call of _addSubscriber()_ method.

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
@@ -7,6 +7,8 @@ import edu.kis.powp.jobs2d.command.ICompoundCommand;
 import edu.kis.powp.observer.Publisher;
 import edu.kis.powp.observer.Subscriber;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 
 /**
@@ -16,6 +18,8 @@ public class DriverCommandManager {
 	private DriverCommand currentCommand = null;
 
 	private Publisher changePublisher = new Publisher();
+
+	private Collection<Subscriber> cache = null;
 
 	/**
 	 * Set current command.
@@ -71,7 +75,12 @@ public class DriverCommandManager {
 	 */
 
 	public synchronized void deleteCurrentObservers(){
+		
+		if(changePublisher.getSubscribers().size() > 0){
+			cache = new ArrayList<>(changePublisher.getSubscribers());
+		}
 
+		changePublisher.clearObservers();
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -56,6 +68,82 @@ public synchronized String getCurrentCommandString() {
...
+	/**
+	 * Add new observer for list of subscribers that belongs to <code>changePublisher</code>
+	 * It should clear cache, if exists.
+	 * @param subscriber new observer to add.
+	 */
+	public synchronized void addObserver(Subscriber subscriber){
+		if(subscriber != null){
+			changePublisher.addSubscriber(subscriber);
+		}
+	}
```
kowallus: The idea was to make the publisher available so we do not need to duplicate its API. However, this is all worth considering.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -2,10 +2,12 @@
 
 import edu.kis.powp.jobs2d.command.DefaultCompoundCommand;
 import edu.kis.powp.jobs2d.command.DriverCommand;
-import edu.kis.powp.jobs2d.command.DriverCommandVisitor;
-import edu.kis.powp.jobs2d.command.ICompoundCommand;
+import edu.kis.powp.jobs2d.drivers.DriverManager;
```
kowallus, line 5: This is another dependency inside DriverCommandManager, but I think it is acceptable.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -56,6 +68,82 @@ public synchronized String getCurrentCommandString() {
...
+
+	/**
+	 * Restores observers from cache
+	 * NOTE:
+	 * It's probable if addObserver/s feature would be implemented, the cache should be empty
+	 */
+
+	public synchronized void resetObservers(){
+		if(cache != null && cache.size() > 0){
+			cache.forEach(cached -> changePublisher.addSubscriber(cached));
```
kowallus: Isn't it a duplication of addObservers method code?

kowallus (approved):
I have some question but OK in general

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
k-chr: Method removed, I came back to previous basic solution in latest commit

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
k-chr: You're right, I changed it to addSubscribers(cache) call instead of that duplication in latest commit.

PR #21, opened 2020-05-12
Importing command from text:

wojciechciejka:
- added option "Open from Json" in "User Test"
- created SelectOpenFromJsonOptionListener class with "actionPerformed" method
- added 3 json files with examples

### DIFF
--- a/iis.json
+++ b/iis.json
@@ -1,0 +1,78 @@
+{
+   "1": {
+         "command": "setPosition",
+		 "x": "-20",
+		 "y": "-50"
+    },
+    "2": {
+         "command": "operateTo",
+		 "x": "-20",
+		 "y": "-50"
+	},
+	"3": {
+         "command": "setPosition",
+		 "x": "-20",
+		"y": "-40"      
+    },
+	"4": {
+         "command": "operateTo",
+		 "x": "-20",
+		 "y": "50" 
+    },
+	"5": {
+         "command": "setPosition",
+		 "x": "0",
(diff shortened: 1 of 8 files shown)


iis.json
```diff
@@ -0,0 +1,78 @@
+{
+   "1": {
```
kowallus, line 2: What are the numbers for? Is it only for preserving the order? If so, you better use json list collection.

src/main/java/edu/kis/powp/jobs2d/events/SelectOpenFromJsonOptionListener.java
```diff
@@ -0,0 +1,68 @@
...
+            for (int i = 0; i < jsonObject.size(); i++) {
+                JSONObject element = (JSONObject) jsonObject.get(String.valueOf(i + 1));
+                String command = (String) element.get("command");
+                int x = Integer.parseInt(String.valueOf(element.get("x")));
+                int y = Integer.parseInt(String.valueOf(element.get("y")));
+                if (command.contains("setPosition"))
+                    commands.add(new SetPositionCommand(x, y));
+                else if (command.contains("operateTo"))
+                    commands.add(new OperateToCommand(x, y));
+
```
kowallus, line 53: Parsing json to command should be reusable! Now it is tangled with e.g. reading the file and the listener. Extract this function.

kowallus (changes requested):
Refactor and minor tuning needed.

PR #24, opened 2020-05-12
IO-1 Extending Command Manager window

Miretpl:
Patryk Janowski
Przemysław Mirowski


src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -68,6 +70,14 @@ public synchronized void clearCurrentCommand() {
         currentCommand = null;
     }
 
+    public synchronized void runCurrentCommand() {
+        currentCommand.execute(driverManager.getCurrentDriver());
+    }
+
+    public synchronized void setCurrentDriver(DriverManager driverManager) {
```
kowallus: setCurrentDriver -> setDriverManager

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -116,5 +115,5 @@ public void run() {
             }
         });
     }
-
+
```
kowallus: only whitespace modifications. Revert this file

kowallus (changes requested):
Nice. Only minor issues.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
raionsakana: Done

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
raionsakana: Done

PR #25, opened 2020-05-12
IO-1: A deep copy of a complex command

arekdziuda:
Arkadiusz Dziuda 215714
Alan Franas 215717

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/DriverCommand.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/DriverCommand.java
@@ -2,10 +2,12 @@ package edu.kis.powp.jobs2d.command;
 
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import java.io.Serializable;
+
 /**
  * DriverCommand interface.
  */
-public interface DriverCommand {
+public interface DriverCommand extends Cloneable, Serializable {
 
     /**
      * Execute command on driver.
@@ -14,4 +16,5 @@ public interface DriverCommand {
      */
     public void execute(Job2dDriver driver);
 
+    public DriverCommand clone();
 }
--- a/src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
@@ -9,13 +9,13 @@ import java.util.List;
  */
 public interface ICompoundCommand extends DriverCommand, Iterable<DriverCommand> {
 
-    public Iterator<DriverCommand> iterator();
+    Iterator<DriverCommand> iterator();
 
     @Override
     default DriverCommand clone(){
         List<DriverCommand> commands = new ArrayList<>();
         for (DriverCommand command : this) {
-            commands.add(command);
+            command.clone();
         }
         return new DefaultCompoundCommand(commands);
     }
--- a/src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
@@ -1,12 +1,23 @@
 package edu.kis.powp.jobs2d.command;
 
+import java.util.ArrayList;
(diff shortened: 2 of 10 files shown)


src/test/java/edu/kis/powp/jobs2d/events/CopyCommandListener.java
```diff
@@ -0,0 +1,39 @@
...
+        commands.add(new OperateToCommand(0, -50));
+        commands.add(new SetPositionCommand(0, -40));
+        commands.add(new OperateToCommand(0, 50));
+        commands.add(new SetPositionCommand(70, -50));
+        commands.add(new OperateToCommand(20, -50));
+        commands.add(new OperateToCommand(20, 0));
+        commands.add(new OperateToCommand(70, 0));
+        commands.add(new OperateToCommand(70, 50));
+        commands.add(new OperateToCommand(20, 50));
+        DriverCommand driverCommand = CommandsFeature.deepCopyDriverCommand(new DefaultCompoundCommand(commands));
```
kowallus, line 34: If you want to test DEEP cloning you must put a compound command inside compound command. Next you can try to modify the original to test if copy changes.

src/main/java/edu/kis/powp/jobs2d/command/DriverCommand.java
```diff
@@ -2,10 +2,12 @@
 
 import edu.kis.powp.jobs2d.Job2dDriver;
 
+import java.io.Serializable;
+
 /**
  * DriverCommand interface.
  */
-public interface DriverCommand {
+public interface DriverCommand extends Cloneable, Serializable {
```
kowallus, line 10: Please, write in comment what is Serializable interface for (in this task)

src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
```diff
@@ -1,12 +1,23 @@
...
-public interface ICompoundCommand extends DriverCommand {
+public interface ICompoundCommand extends DriverCommand, Iterable<DriverCommand> {
 
     public Iterator<DriverCommand> iterator();
 
+    @Override
+    default DriverCommand clone(){
+        List<DriverCommand> commands = new ArrayList<>();
+        for (DriverCommand command : this) {
+            commands.add(command);
```
kowallus: Shouldn't you use command.clone() here?

kowallus (changes requested):
Quite nice. Some issues and questions.

src/main/java/edu/kis/powp/jobs2d/command/DriverCommand.java
DeanMiles: I have used this in 30 line in  CommandFeature.deepCopyDriverCommand

> return SerializationUtils.clone(driverCommandToCopy);

I had to implement Serializable to use SerializationUtils because clone in this library as parametr has T object

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/SerializationUtils.html

src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
arekdziuda: Fixed in next commit

src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
```diff
@@ -9,13 +9,13 @@
...
 
-    public Iterator<DriverCommand> iterator();
+    Iterator<DriverCommand> iterator();
 
     @Override
     default DriverCommand clone(){
         List<DriverCommand> commands = new ArrayList<>();
         for (DriverCommand command : this) {
-            commands.add(command);
+            command.clone();
```
kowallus, line 18: Does it work now? I don't think so. Has it been tested?

kowallus (changes requested):

kowallus:
Reopen PR after modifications

PR #26, opened 2020-05-13
Importing command from text added.

Grzegorian:
Owocna noc :D


src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -78,4 +85,26 @@ public synchronized String getCurrentCommandString() {
...
+        }.getType();
+        List<Command> all = gson.fromJson(loadedCommands, dataListType);
+        List<DriverCommand> po = new ArrayList<DriverCommand>();
+        for (Command t : all) {
+            if (t.operation.equals("SetPositionCommand")) {
+                po.add(new SetPositionCommand(t.posX, t.posY));
+            } else if (t.operation.equals("OperateToCommand")) {
+                po.add(new OperateToCommand(t.posX, t.posY));
+            }
+        }
```
kowallus, line 107: Parsing should be extracted (it should not be tangled with command management) to separate class.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -78,4 +85,26 @@ public synchronized String getCurrentCommandString() {
...
+        setCurrentCommand(po, "Loaded input");
```
kowallus, line 108: Can you add an optional command name to the input text.

kowallus (changes requested):
Almost OK.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -1,13 +1,20 @@
 package edu.kis.powp.jobs2d.command.manager;
 
-import java.util.Iterator;
-import java.util.List;
-
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
```
kowallus, line 4: What about dependencies to gson? I cannot see pom.xml changes

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
Grzegorian: Can you write me an example?

kowallus: Something like:
{
"name":  "a point",
"commands": [
  {
    "operation": "SetPositionCommand",
    "posX": -20,
    "posY": -50
  },
  {
    "operation": "OperateToCommand",
    "posX": -20,
    "posY": -50
  }
] 
}

kowallus:
Reopen PR after modifications

PR #30, opened 2020-05-13
IO-3 MACRO - recording driver calls as commands v2

AndrzejBilant:


src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverAdapter.java
```diff
@@ -0,0 +1,55 @@
...
+import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.command.DriverCommand;
+import edu.kis.powp.jobs2d.command.OperateToCommand;
+import edu.kis.powp.jobs2d.command.SetPositionCommand;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class MacroDriverAdapter implements Job2dDriver {
```
kowallus, line 12: Not right. This is decorator!

src/test/java/edu/kis/powp/jobs2d/events/SelectLoadMacroListener.java
```diff
@@ -0,0 +1,26 @@
...
+    public SelectLoadMacroListener(DriverManager driverManager) {
+        this.driverManager = driverManager;
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        CommandsFeature.getDriverCommandManager().setCurrentCommand(MacroFeature.getMacroDriverAdapter().getDriverCommandList(), "Macro");
+        driverManager.setCurrentDriver(MacroFeature.getMacroDriverAdapter().getDriver());
+        DriverCommand driverCommand = CommandsFeature.getDriverCommandManager().getCurrentCommand();
+        driverCommand.execute(driverManager.getCurrentDriver());
```
kowallus, line 23: If name changes from 'run' to 'load' then the logic should also be updated. Please remove it and test if it works using "Run command" from menu.

kowallus (changes requested):
Closer, but not enough.

kowallus:
Reopen this PR, after modifications.

BartoszKowalczyk98:
We don't seem to have an option to reopen our PR.

PR #33, opened 2020-05-16
Immutable complex command

michaaeu:


src/main/java/edu/kis/powp/jobs2d/command/ImmutableComplexCommand.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.List;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class ImmutableComplexCommand implements ICompoundCommand {
+    private List<DriverCommand> driverCommandList;
+
+    public ImmutableComplexCommand(List<DriverCommand> driverCommandList) {
+        this.driverCommandList = Collections.unmodifiableList(driverCommandList);
+    }
```
kowallus, line 14: Note that there might be a problem if you will get a mutable complex command as an element of a driverCommandList since it still can be modified from outside.

kowallus (approved):
I approve, however, the solution is not 100% solid. It may be easier to fix it later (when other tasks will be finished).

src/main/java/edu/kis/powp/jobs2d/command/ImmutableComplexCommand.java
michaaeu: Update in #106

PR #34, opened 2020-05-16
MACRO IO2

RafalProoo:


src/main/java/edu/kis/powp/jobs2d/drivers/MacroDriverDecorator.java
```diff
@@ -0,0 +1,45 @@
...
+import edu.kis.powp.jobs2d.command.SetPositionCommand;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class MacroDriverDecorator implements Job2dDriver {
+    private List<DriverCommand> driverCommandList = new ArrayList<>();
+    private Job2dDriver job2dDriver;
+
+    public void setJob2dDriver(Job2dDriver job2dDriver) {
```
kowallus, line 15: setCoreJob2bDriver would be more precise (the same with getter)

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -18,6 +13,12 @@
 import edu.kis.powp.jobs2d.features.CommandsFeature;
 import edu.kis.powp.jobs2d.features.DrawerFeature;
 import edu.kis.powp.jobs2d.features.DriverFeature;
+import edu.kis.powp.jobs2d.features.MacroFeature;
+
+import java.awt.*;
+import java.awt.event.ActionEvent;
+import java.util.logging.Level;
+import java.util.logging.Logger;
```
kowallus, line 21: please revert order of imports

kowallus (changes requested):
Listeners are not used anywhere!? Did you incorrectly copy/paste solutions from other groups and even didn't test it?!

kowallus:
Reopen after modifications

PR #36, opened 2020-05-17
Add visitor in commands hierarchy and tests

tobiaszkubiak31:
Łukasz Zalewski 215931, Tobiasz Kubiak 215784

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/CommandCounterVisitor.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/CommandCounterVisitor.java
@@ -4,7 +4,6 @@ import java.util.Iterator;
 import java.util.logging.Logger;
 
 public class CommandCounterVisitor implements Visitor {
-    private final static Logger logger = Logger.getLogger(Logger.GLOBAL_LOGGER_NAME);
 
     private int operateToCommandCounter = 0;
     private int setPositionCommandCounter = 0;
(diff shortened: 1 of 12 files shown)


src/main/java/edu/kis/powp/jobs2d/command/CommandCounterVisitor.java
```diff
@@ -0,0 +1,42 @@
...
+    public void visit(SetPositionCommand setPositionCommand) {
+        setPositionCommandCounter++;
+    }
+
+    @Override
+    public void visit(ICompoundCommand driverCommand) {
+        Iterator<DriverCommand> iterator = driverCommand.iterator();
+        while (iterator.hasNext())
+        {
+            DriverCommand driverCommand1 = iterator.next();
```
kowallus, line 38: I don't like '1' suffix (it has no sense)
My proposition of names: innerCommand or simply c. 
You can also change the name of param to compoundCommand to make code coherent

src/main/java/edu/kis/powp/jobs2d/command/SetPositionCommand.java
```diff
@@ -20,4 +20,8 @@ public void execute(Job2dDriver driver) {
 		driver.setPosition(posX, posY);
 	}
 
+	@Override
+	public void accept(Visitor visitor) {
+		visitor.visit(this);
+	}
```
kowallus, line 26: You can add a similar implementation to ICompoundCommand (default implementation)

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -28,17 +30,25 @@ public synchronized void setCurrentCommand(DriverCommand commandList) {
...
-	 * 
+	 *
 	 * @param commandList list of commands representing a compound command.
 	 * @param name        name of the command.
 	 */
 	public synchronized void setCurrentCommand(List<DriverCommand> commandList, String name) {
 		setCurrentCommand(new ICompoundCommand() {
 
 			List<DriverCommand> driverCommands = commandList;
+			CommandCounterVisitor commandCounterVisitor = new CommandCounterVisitor();
```
kowallus, line 41: NO. You should not add a particular visitor here!

src/test/java/edu/kis/powp/jobs2d/ICompoundCommandVisitorTest.java
```diff
@@ -0,0 +1,92 @@
...
+            String name;
+
+            @Override
+            public void accept(Visitor visitor) {
+                visitor.visit(this);
+            }
+
+            @Override
+            public void execute(Job2dDriver driver) {
+                driverCommands.forEach((c) -> c.accept(commandCounter));
```
kowallus, line 53: must be removed with commandCounter

kowallus (changes requested):
Not bad, but it requires several simplifications.

kowallus:
Reopen after fixes.

PR #45, opened 2020-05-18
Added an implenentation of ImmutableComplexCommand

215848:
Co-authored-by: 215848 <215848@edu.p.lodz.pl>


kowallus (approved):
OK, however, the solution is not 100% solid (and not tested). There will be a problem when a complex command will contain a mutable complex command. We can try to fix it later (after other tasks will be completed).

One thing: the convention of the project uses tabs instead of spaces. Fix before merge.

215848:
Fixed indentations

PR #46, opened 2020-05-18
Io 3 Convert mouse clicks to driver calls

MR-98:
Mateusz Rosiak 215866
Mateusz Łukomski 215805


src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -114,7 +113,10 @@ public void run() {
 				setupLogger(app);
 				setupWindows(app);
 
-				app.setVisibility(true);
+                JPanel panel = app.getFreePanel();
+                panel.addMouseListener(new MouseDrawListener(panel));
+
+                app.setVisibility(true);
```
kowallus: the original code convention uses tabs instead of spaces! fix it

src/main/java/edu/kis/powp/jobs2d/events/MouseDrawListener.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private JPanel panel;
+
+    public MouseDrawListener(JPanel panel) {
+        this.panel = panel;
+    }
+
+    @Override
+    public void mouseClicked(MouseEvent e) {
+        Job2dDriver currentDriver = DriverFeature.getDriverManager().getCurrentDriver();
```
kowallus: Avoid a dependency to DriverFeature. You can just keep a reference to the manager.

kowallus (changes requested):
OK, but minor issues require attention. Please close PR and reopen after modifications.

PR #47, opened 2020-05-18
Io-3 Extending Command Manager window

zenakz:
David Zientak 215934
Jędrzej Lis-Marciniak 215799

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,8 +13,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.0</version>
 				<configuration>
-					<source>1.8</source>
-					<target>1.8</target>
+					<source>10</source>
+					<target>10</target>
 					<encoding>UTF-8</encoding>
 					<showDeprecation>true</showDeprecation>
 				</configuration>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -13,8 +13,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.0</version>
 				<configuration>
-					<source>1.8</source>
-					<target>1.8</target>
+					<source>10</source>
+					<target>10</target>
```
kowallus: Is it really used?

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -65,26 +66,62 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
 	}
 
 	private void clearCommand() {
 		commandManager.clearCurrentCommand();
 		updateCurrentCommandField();
 	}
 
+	private void runCommand() {
+		commandManager.runCurrentCommand();
+		updateCurrentCommandField();
```
kowallus: I think that this line is not necessary. Can you verify it?

kowallus (approved):
OK. Please verify pointed out issues before merging.

kowallus:
I hope that the last commit has been tested ;)

PR #51, opened 2020-05-19
Io 2 dev

Mmati1996:


src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -2,20 +2,23 @@
...
 import edu.kis.powp.jobs2d.LoggerDriver;
+import edu.kis.powp.observer.Publisher;
 
 /**
  * Driver manager provides means to setup the driver. It also enables other
  * components and features of the application to react on configuration changes.
  */
 public class DriverManager {
 
+	Publisher publisher = new Publisher();
```
kowallus: Should be private!

kowallus (approved):
I approve and merge. Only one thing. Prepare in the separate PR.

kowallus:
There are conflicts. Resolve conflicts & add private to publisher field.

kowallus:
Conflicts with the main branch still not resolved!

PR #55, opened 2020-05-20
Monitoring usage of a device IO1

MichalOgloza:
Michał Ogłoza 215834
Krzysztof Chybicki 215693

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,8 +14,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.0</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>10</source>
+                    <target>10</target>
                     <encoding>UTF-8</encoding>
                     <showDeprecation>true</showDeprecation>
                 </configuration>
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
@@ -10,12 +10,14 @@ import edu.kis.powp.jobs2d.LoggerDriver;
 public class DriverManager {
 
     private Job2dDriver currentDriver = new LoggerDriver();
+    private DriverStatistics currentStatistics = new DriverStatistics();
 
     /**
      * @param driver Set the driver as current.
      */
-    public synchronized void setCurrentDriver(Job2dDriver driver) {
+    public synchronized void setCurrentDriver(Job2dDriver driver, DriverStatistics statistics) {
         currentDriver = driver;
+        currentStatistics = statistics;
     }
 
     /**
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -14,8 +14,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.0</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>10</source>
+                    <target>10</target>
```
kowallus, line 18: I would prefer staying with 1.8. Is it really so important

src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -10,12 +10,14 @@
 public class DriverManager {
 
     private Job2dDriver currentDriver = new LoggerDriver();
+    private DriverStatistics currentStatistics = new DriverStatistics();
 
     /**
      * @param driver Set the driver as current.
      */
-    public synchronized void setCurrentDriver(Job2dDriver driver) {
+    public synchronized void setCurrentDriver(Job2dDriver driver, DriverStatistics statistics) {
```
kowallus, line 18: Please do not force operating on drivers together with statistics (mixing responsibilities here). When "Observer for a driver changes events" task will be finished you will be able to achieve similar functionality without tangling the implementation.

src/main/java/edu/kis/powp/jobs2d/drivers/SelectDriverMenuOptionListener.java
```diff
@@ -9,15 +9,17 @@
...
+    public SelectDriverMenuOptionListener(Job2dDriver driver, DriverManager driverManager, DriverStatistics statistics) {
         this.driverManager = driverManager;
         this.driver = driver;
+        this.statistics = statistics;
     }
 
     @Override
     public void actionPerformed(ActionEvent e) {
-        driverManager.setCurrentDriver(driver);
+        driverManager.setCurrentDriver(driver, statistics);
```
kowallus, line 22: Temporarily use the separate option to set stats.

src/main/java/edu/kis/powp/jobs2d/drivers/adapter/LineDriverAdapter.java
```diff
@@ -11,31 +15,40 @@
     private ILine line;
     private int startX = 0, startY = 0;
     private String name;
+    private DriverStatistics statistics;
 
     private DrawPanelController drawController;
 
-    public LineDriverAdapter(DrawPanelController drawController, ILine line, String name) {
+    public LineDriverAdapter(DrawPanelController drawController, ILine line, String name, DriverStatistics statistics) {
```
kowallus, line 22: Statistics should be separate from drivers (SRP principle!)

kowallus (changes requested):
Not satisfying design. Please, rework the design of the solution.

kowallus:
Please close PR and reopen after modifications.

PR #58, opened 2020-05-22
Io 1 dev

DeanMiles:


src/main/java/edu/kis/powp/jobs2d/command/OperateToCommand.java
```diff
@@ -20,4 +20,15 @@ public void execute(Job2dDriver driver) {
...
     }
 
+    @Override
+    public DriverCommand clone(){
+        DriverCommand dc = null;
+        try {
+            dc = (DriverCommand) super.clone();
+        } catch (CloneNotSupportedException e) {
+            e.printStackTrace();
+        }
```
kowallus, line 30: Is there a real risk of an exception here? I would resign from catch block here.

src/test/java/edu/kis/powp/jobs2d/events/CopyCommandListener.java
```diff
@@ -0,0 +1,41 @@
...
+        commands.add(new SetPositionCommand(0, -50));
+        commands.add(new OperateToCommand(0, -50));
+        commands.add(new SetPositionCommand(0, -40));
+        commands.add(new OperateToCommand(0, 50));
+        commands.add(new SetPositionCommand(70, -50));
+        commands.add(new OperateToCommand(20, -50));
+        commands.add(new OperateToCommand(20, 0));
+        commands.add(new OperateToCommand(70, 0));
+        commands.add(new OperateToCommand(70, 50));
+        commands.add(new OperateToCommand(20, 50));
```
kowallus, line 33: This could be extracted to some factory for potential re-use.

kowallus (approved):
Only some comments

PR #59, opened 2020-05-22
Usable commands factory

Luvkitri:
Basic draft for command factory.


startled-cat:
We could use some feedback and more specific information about what does "manage and catalogue commands" mean.

kowallus:
The topic is more about ability to manage a collection of various complex commands (e.g. representing different shapes or more complex jobs). "manage and catalogue" should make it easier for users to find commands depending on a need (some labels, keywords e.g.). It should be possible to add a complex command to the factory from a loaded command (setCurrentCommand)

PR #61, opened 2020-05-25
Monitoring usage of a device IO4

asmei1:


src/main/java/edu/kis/powp/jobs2d/command/gui/MonitorUsageWindow.java
```diff
@@ -0,0 +1,65 @@
...
+		container.add(scrollPane,  BorderLayout.CENTER);
+	}
+	
+	@Override
+	public void HideIfVisibleAndShowIfHidden() {
+		if (this.isVisible()) {
+			this.setVisible(false);
+		} else {
+			this.setVisible(true);
+		}
```
kowallus: this.setVisible(!this.isVisible())
:)

src/main/java/edu/kis/powp/jobs2d/command/gui/MonitorUsageWindow.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+	
+	@Override
+	public void update() {
+		this.setTitle("Monitor usage - " + DriverFeature.getDriverManager().getCurrentDriver().toString());
```
kowallus: Please try to remove these dependencies DriverFeature.getDriverManager().getCurrentDriver().toString(). 
This window does not need to be a subscriber (can be extracted to separate class) and it only needs to know the name of the driver.

src/main/java/edu/kis/powp/jobs2d/drivers/DriverUsageMonitor.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.logging.Logger;
+
+public class DriverUsageMonitor extends Job2dDriverDecorator {
+	Logger logger = Logger.getLogger("usagemonitor");
+	
+	private int operateToDistance;
+	private int setPositionDistance;
+	private int previousPositionX;
+	private int previousPositionY;
```
kowallus: please provide initial values

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -32,65 +34,65 @@ private static void setupPresetTests(Application application) {
...
+	
 	/**
 	 * Setup driver manager, and set default Job2dDriver for application.
 	 *
 	 * @param application Application context.
 	 */
 	private static void setupDrivers(Application application) {
-		Job2dDriver loggerDriver = new LoggerDriver();
-
+		Job2dDriver loggerDriver = new DriverUsageMonitor(new LoggerDriver());
```
kowallus: You should not force the decorator in that way. If anyone else adds a new driver your monitor will not work. You can add your own driver combination without changing existing ones.
LATER you can provide some generic mechanism, taking the advantage from "Observer for a driver changes events" task.

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -32,65 +34,65 @@ private static void setupPresetTests(Application application) {
...
+		
 		DrawPanelController drawerController = DrawerFeature.getDrawerController();
-		Job2dDriver driver = new LineDriverAdapter(drawerController, LineFactory.getBasicLine(), "basic");
+		Job2dDriver driver = new DriverUsageMonitor(new LineDriverAdapter(drawerController, LineFactory.getBasicLine(), "basic"));
 		DriverFeature.addDriver("Line Simulator", driver);
 		DriverFeature.getDriverManager().setCurrentDriver(driver);
-
-		driver = new LineDriverAdapter(drawerController, LineFactory.getSpecialLine(), "special");
+		
+		driver = new DriverUsageMonitor(new LineDriverAdapter(drawerController, LineFactory.getSpecialLine(), "special"));
```
kowallus: the same issue as above

kowallus (changes requested):
Almost ok (for the first stage of monitoring). Consider the following comments.

kowallus:
Please close PR and reopen after the update.

src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -17,7 +17,7 @@
 	 * @param driver Set the driver as current.
 	 */
 	public synchronized void setCurrentDriver(Job2dDriver driver) {
-		currentDriver = driver;
+		currentDriver = driver instanceof DriverUsageMonitor ? driver : new DriverUsageMonitor(driver);
```
kowallus: Do not change this code. Try decorating in a subscriber (this way usage monitor would be optional, not hardcoded)

kowallus (changes requested):
One important issue

kowallus:
Please close and reopen after fixes

kowallus:
Further sub-tasks added to trello in checklist.

PR #63, opened 2020-05-25
IO-4 History of used commands

Trackres:


src/main/java/edu/kis/powp/jobs2d/command/gui/backlog/BackLogManagerWindowCommandChangeObserver.java
```diff
@@ -0,0 +1,17 @@
+package edu.kis.powp.jobs2d.command.gui.backlog;
+import edu.kis.powp.observer.Subscriber;
+
+public class BackLogManagerWindowCommandChangeObserver implements Subscriber {
+
+    private BackLogManagerWindow backLogManagerWindow;
```
kowallus, line 6: Try to redesign the history component to be independent of the GUI. (e.g. we might want to change GUI or use history from another component without accessing GUI).
There should be a separate class representing the history (it makes more sense if history would be more than a text - returning to previous commands).

kowallus (changes requested):
One important suggestion

PR #64, opened 2020-05-26
Feature Command (device) usage analyser

k-chr:


src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -53,9 +56,18 @@ public synchronized DriverCommand getCurrentCommand() {
 	 */
 	public synchronized void setCurrentCommand(DriverCommand commandList) {
 		this.currentCommand = commandList;
+
+		if(this.currentCommand instanceof ICompoundCommand) {
+			analyzer.analyze((ICompoundCommand) this.currentCommand);
+		}
```
kowallus: reimplement using observer. This code can be in a separate subscriber class registred in changePublisher

src/main/java/edu/kis/powp/jobs2d/command/analyzer/ICommandUsageAnalyzer.java
```diff
@@ -0,0 +1,28 @@
+package edu.kis.powp.jobs2d.command.analyzer;
+
+import edu.kis.powp.jobs2d.command.DriverCommandVisitor;
+import edu.kis.powp.jobs2d.command.ICompoundCommand;
+
+/**
+ * This interface was defined for complex commands benchmarking and analysis of usage purpose.
+ */
+public interface ICommandUsageAnalyzer extends DriverCommandVisitor {
```
kowallus: Verify if DriverCommandVisitor needs to be a part of this interface

kowallus (changes requested):
Let's start with two issues. The solution is too large to check in few minutes.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
k-chr: ok, I wondered about this approach earlier, I will implement it ASAP

k-chr: updated in latest hotfix

src/main/java/edu/kis/powp/jobs2d/command/analyzer/ICommandUsageAnalyzer.java
k-chr: updated in latest hotfix

src/main/java/edu/kis/powp/jobs2d/command/analyzer/StandardComputationPolicy.java
```diff
@@ -0,0 +1,139 @@
...
+        double factor = (deviceDistanceFactor / (deviceDistanceFactor - 1));
+        double time = Math.sqrt(coefficient / deviceAcceleration) * factor * writeFactor;
+        double averageVelocity = distance / time;
+        averageVelocity = Double.isNaN(averageVelocity) ? 0 : averageVelocity;
+        double totalInk = type == UsageType.WRITE ? distance / inkFactor : 0;
+
+        statistics.addRecord("time", time);
+        statistics.addRecord("averageVelocity", averageVelocity);
+        statistics.addRecord("distance", distance);
+        statistics.addRecord("totalInkUsed", totalInk);
```
kowallus: you should use enum or at least constants for those strings

src/main/java/edu/kis/powp/jobs2d/command/analyzer/CommandUsageAnalyzerImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        visitor.setStartPosition(startX, startY);
+        visitor.visit(compoundCommand);
+
+        LinkedList<Usage> usagesOfHead = visitor.getUsages();
+        usagesOfHead.forEach(usage -> {
+            Statistics statistics = policy.compute(usage.getStartX(), usage.getStartY(), usage.getEndX(), usage.getEndY(), usage.getType());
+            this.timeOfUsages.add(statistics.getData("time"));
+            this.averageVelocities.add(statistics.getData("averageVelocity"));
+            this.distances.add(statistics.getData("distance"));
+            this.inkUsages.add(statistics.getData("totalInkUsed"));
```
kowallus: enum or constants

src/main/java/edu/kis/powp/jobs2d/command/analyzer/CommandAnalyzerCommandChangeObserver.java
```diff
@@ -0,0 +1,26 @@
...
+        this.analyzer = analyzer;
+    }
+
+
+    @Override
+    public void update() {
+        DriverCommand command = CommandsFeature.getDriverCommandManager().getCurrentCommand();
+        if(command instanceof ICompoundCommand){
+            analyzer.analyze((ICompoundCommand) command);
+        }
```
kowallus: Why is it a problem to run on other kind of DriverCommand? 
Try to avoid instanceof (visitor makes it possible)

src/main/java/edu/kis/powp/jobs2d/command/analyzer/CommandUsageAnalyzerImpl.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public void analyze(ICompoundCommand compoundCommand) {
+
+        timeOfUsages = new LinkedList<>();
+        distances = new LinkedList<>();
+        averageVelocities = new LinkedList<>();
+        inkUsages = new LinkedList<>();
+
+        visitor.setStartPosition(startX, startY);
+        visitor.visit(compoundCommand);
```
kowallus: you should not use visit directly (always use accept)

src/main/java/edu/kis/powp/jobs2d/command/analyzer/StandardComputationPolicy.java
```diff
@@ -0,0 +1,139 @@
...
+        return statistics;
+    }
+
+    private double mapUnit(Unit distanceUnit) {
```
kowallus: maybe it could be converted to public static utility method in Unit?

kowallus (changes requested):
Final issues in comments

kowallus (approved):
OK. I think that is enough. Prepare yourself for a presentation.

k-chr:
> 
> 
> OK. I think that is enough. Prepare yourself for a presentation.

Ok, there' s the only thing I want to know, when presentation will be scheduled for?

PR #65, opened 2020-05-26
dev io4 Commands transformation v2.0

Hergoln:


kowallus (approved):
Better. However, you could provide a reusable hierarchy of coord transformation (it would be a kind of bridge pattern or strategy refactor). We can make it as another task.

PR #67, opened 2020-05-26
Observer for a driver changes events IO1

Kacpri:


src/main/java/edu/kis/powp/jobs2d/drivers/DriverChangeObserver.java
```diff
@@ -0,0 +1,16 @@
+package edu.kis.powp.jobs2d.drivers;
+
+import edu.kis.powp.jobs2d.features.DriverFeature;
+import edu.kis.powp.observer.Subscriber;
+
+
+public class DriverChangeObserver implements Subscriber {
```
kowallus: The name of the observer should explain its behavior: i.e.  DriverFeature.updateDriverInfo();

kowallus (changes requested):
Please fix the name of the observer!

Kacpri:
I've changed the name of the class

PR #68, opened 2020-05-27
Io 1 dev (history of used commands)

arekdziuda:
Arkadiusz Dziuda 215714
Alan Franas 215717

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryController.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryController.java
@@ -1,0 +1,19 @@
+package edu.kis.powp.jobs2d.command.historyComponent;
+
+import edu.kis.powp.appbase.Application;
+import edu.kis.powp.jobs2d.features.CommandsFeature;
+
+import java.awt.event.ActionEvent;
+
+public class HistoryController {
+
+    public static void setupHistoryController(Application application) {
+        HistoryManagerWindow history = new HistoryManagerWindow(CommandsFeature.getDriverCommandManager());
+        HistoryWindowCommandChangeObserver historyObserver = new HistoryWindowCommandChangeObserver(history);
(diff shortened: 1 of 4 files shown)


src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryController.java
```diff
@@ -0,0 +1,19 @@
+package edu.kis.powp.jobs2d.command.historyComponent;
+
+import edu.kis.powp.appbase.Application;
+import edu.kis.powp.jobs2d.features.CommandsFeature;
+
+import java.awt.event.ActionEvent;
+
+public class HistoryController {
```
kowallus, line 8: Maybe HistoryFeature. It does not look like a controller to me.

src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryWindowCommandChangeObserver.java
```diff
@@ -0,0 +1,17 @@
...
+
+    private HistoryManagerWindow backLogManagerWindow;
+    public HistoryWindowCommandChangeObserver(HistoryManagerWindow commandManagerWindow) {
+        super();
+        this.backLogManagerWindow = commandManagerWindow;
+    }
+
+    @Override
+    public void update() {
+        backLogManagerWindow.updateCommandListField();
```
kowallus, line 15: Try to redesign the history component to be independent of the GUI. (e.g. we might want to change GUI or use history from another component without accessing GUI).
There should be a separate class representing the history (it makes more sense if history would be more than a text - returning to previous commands).

kowallus (changes requested):
Some suggestions in comments. 
Please add a feature to be able to return to the given command from the history.

kowallus:
Please close, update and re-open PR

PR #69, opened 2020-05-27
Feature/importing command from text

TheDamik34:


src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -68,4 +69,19 @@ public synchronized String getCurrentCommandString() {
     public Publisher getChangePublisher() {
         return changePublisher;
     }
+
+    static class ComplexCommand {
```
kowallus: The name of this class is confusing with DriverCommand hierarchy and it should be defined elsewhere (changes to DriverCommandManager should be minimal). Is it used in many places?

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -68,4 +69,19 @@ public synchronized String getCurrentCommandString() {
...
+        String name;
+        static class Command {
+            String operation;
+            int posX;
+            int posY;
+        }
+        List<Command> commands;
+    }
+
+    public synchronized void loadCommands(String loadedCommands) {
```
kowallus: The name should be e.g. loadCommandsFromJson. This is a very simple method, I don't think it is really necessary.

src/main/java/edu/kis/powp/jobs2d/command/manager/JSONCommandParser.java
```diff
@@ -0,0 +1,38 @@
...
+import edu.kis.powp.jobs2d.command.OperateToCommand;
+import edu.kis.powp.jobs2d.command.SetPositionCommand;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class JSONCommandParser {
+    private DriverCommandManager.ComplexCommand complexCommandFromJSON;
+
+    public JSONCommandParser(String json) {
```
kowallus: parser should be reusable. There should be a separate method for parsing (then you can should think also about an interface, to enable implentation of various parsers in the future)

src/test/java/edu/kis/powp/jobs2d/events/SelectLoadSecretCommandOptionListener.java
```diff
@@ -1,16 +1,16 @@
...
 import edu.kis.powp.jobs2d.command.OperateToCommand;
 import edu.kis.powp.jobs2d.command.SetPositionCommand;
 import edu.kis.powp.jobs2d.command.manager.DriverCommandManager;
 import edu.kis.powp.jobs2d.features.CommandsFeature;
 
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.util.ArrayList;
+import java.util.List;
+
```
kowallus: This class has only order of imports changed! Revert!

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -59,6 +55,18 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
 
+        InputCommandsTextArea = new JTextArea("");
+        InputCommandsTextArea.setEditable(true);
+        InputCommandsTextArea.setBorder(BorderFactory.createLineBorder(Color.BLUE));
+        InputCommandsTextArea.setLineWrap(true);
+
+        JScrollPane InputCommandsField = new JScrollPane(InputCommandsTextArea);
+        content.add(InputCommandsField,c);
+
+        JButton btnLoadCommand = new JButton("Load commands");
```
kowallus: The name should suggest JSON input.

kowallus (changes requested):
Better, but there are still small issues to improve.

kowallus:
Please close, fix and reopen the PR

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -97,6 +106,17 @@ public void handleObservers(JButton button) {
         this.isDeleted = !this.isDeleted;
     }
 
+    private void loadCommandsFromJSON() {
+        String jsonInput = InputCommandsTextArea.getText().trim();
+        JSONCommandParser jsonCommandParser = new JSONCommandParser(jsonInput);
+        jsonCommandParser.parseToDriverCommand();
```
kowallus: jsonInput should be a parameter of this method to re-use parser. A jsonCommandParser instance could be then re-used (stored in the CommandManagerWindow instance as an attribute)

src/main/java/edu/kis/powp/jobs2d/command/manager/parsers/JSONCommandParser.java
```diff
@@ -0,0 +1,58 @@
...
+
+        complexCommandFromJSON =
+                gson.fromJson(json, DriverCommandAdapterJSON.class);
+    }
+
+    public List<DriverCommand> getDriverCommand() {
+        List<DriverCommand> complexCommand = new ArrayList<>();
+
+        complexCommandFromJSON.commands.forEach(command -> {
+            if(command.operation.equals("SetPositionCommand")){
```
kowallus: Constants should be extracted.

src/main/java/edu/kis/powp/jobs2d/command/manager/parsers/JSONCommandParser.java
```diff
@@ -0,0 +1,58 @@
...
+    public void setJson(String json) {
+        this.json = json;
+    }
+
+    @Override
+    public void parseToDriverCommand() {
+        Gson gson = new Gson();
+
+        complexCommandFromJSON =
+                gson.fromJson(json, DriverCommandAdapterJSON.class);
```
kowallus: json is actually used only here, so it could be a param of this method

src/main/java/edu/kis/powp/jobs2d/command/manager/parsers/Parser.java
```diff
@@ -0,0 +1,5 @@
+package edu.kis.powp.jobs2d.command.manager.parsers;
+
+public interface Parser {
+    void parseToDriverCommand();
```
kowallus: This interface does not make any sense. The client can only parse and cannot get anything in return!?

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -97,6 +106,17 @@ public void handleObservers(JButton button) {
         this.isDeleted = !this.isDeleted;
     }
 
+    private void loadCommandsFromJSON() {
+        String jsonInput = InputCommandsTextArea.getText().trim();
+        JSONCommandParser jsonCommandParser = new JSONCommandParser(jsonInput);
```
kowallus: You should be able to use parser using only Parser interface! If not, this interface has no value.

src/main/java/edu/kis/powp/jobs2d/command/manager/parsers/JSONCommandParser.java
```diff
@@ -0,0 +1,58 @@
...
+public class JSONCommandParser implements Parser {
+    private static class DriverCommandAdapterJSON {
+        String name;
+        static class Command {
+            String operation;
+            int posX;
+            int posY;
+        }
+        List<Command> commands;
+    }
```
kowallus: It is ok, but just a thought. This class does not seem to be tight strongly with json. Maybe it could be reused in other (potential future) parsers?

kowallus (changes requested):
Still some issues. The Parser interface is not really reusable.

kowallus:
please close, update and reopen.

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -82,14 +85,12 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
         content.add(btnClearObservers, c);
     }
 
-    private void loadCommandsFromJSON() {
-        String jsonInput = InputCommandsTextArea.getText().trim();
-        JSONCommandParser jsonCommandParser = new JSONCommandParser(jsonInput);
-        jsonCommandParser.parseToDriverCommand();
+    private void loadCommandsFromJSON(String jsonInput) {
+        InputDataModel inputDataModel = jsonCommandParser.parse(jsonInput);
```
kowallus, line 124: Just a comment: the change seems irrelevant, but now it should be easier, to actually extract this outside GUI.

kowallus (approved):
It might be more simple, but OK.

PR #70, opened 2020-05-29
IO4 - Usable commands factory

startled-cat:


src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public ICompoundCommand get(String name){
+        return shapes.get(name);
+    }
+
+    public void add(String name, ICompoundCommand commands){
+        List<DriverCommand> list = new ArrayList<>();
+        for (DriverCommand dc : commands) {
+            list.add(dc);
+        }
```
kowallus, line 28: This is a shallow copy. Better re-use "A deep copy of a complex command".

kowallus (changes requested):
You have provided a simple factory core component. OK, but it is not enough yet for merging. There should be at least a minimal GUI for inserting and retrieving commands (to/from DriverCommandsManager).

PR #75, opened 2020-06-02
Io 2 dev History of used commands

GGuinea:
215933
215851
215671
215793

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
@@ -1,5 +1,21 @@
 package edu.kis.powp.jobs2d.command.history;
 
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+
+import javax.swing.DefaultListModel;
+
 public class CommandHistory {
 
+    static DefaultListModel<HistoryEntry> entryHistoryList = new DefaultListModel<>();
+
+    public static void addCommandEntry(String commandName) {
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("HH:mm:ss");
+        HistoryEntry historyEntry = new HistoryEntry(LocalTime.now().format(formatter), commandName);
+        entryHistoryList.addElement(historyEntry);
+    }
+
+    public static void clearHistory() {
+        entryHistoryList.clear();
+    }
 }
--- a/src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
@@ -1,0 +1,5 @@
+package edu.kis.powp.jobs2d.command.history;
+
+public class CommandHistory {
+
+}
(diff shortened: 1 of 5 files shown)


src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
```diff
@@ -0,0 +1,21 @@
+package edu.kis.powp.jobs2d.command.history;
+
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+
+import javax.swing.DefaultListModel;
+
+public class CommandHistory {
+
+    static DefaultListModel<HistoryEntry> entryHistoryList = new DefaultListModel<>();
```
kowallus, line 10: It would be safer to provide a read-only view through the getter

src/main/java/edu/kis/powp/jobs2d/command/history/HistoryEntry.java
```diff
@@ -0,0 +1,21 @@
+package edu.kis.powp.jobs2d.command.history;
+
+public class HistoryEntry {
+
+    private String date;
+    private String commandName;
```
kowallus, line 6: There lacks data to reinstate a previous command (it looks more like a log, but I expected an active history). I would like such an option to be implemented also from the GUI.

src/main/java/edu/kis/powp/jobs2d/command/history/HistoryViewer.java
```diff
@@ -0,0 +1,44 @@
...
+
+import edu.kis.powp.appbase.gui.WindowComponent;
+
+public class HistoryViewer extends JFrame implements WindowComponent {
+
+    private static final long serialVersionUID = 1L;
+    private JList<HistoryEntry> historyCommandList;
+
+    public HistoryViewer() throws HeadlessException {
+        historyCommandList = new JList<HistoryEntry>(CommandHistory.entryHistoryList);
```
kowallus, line 21: this parameter could be injected to remove the dependency to CommandHistory

src/main/java/edu/kis/powp/jobs2d/command/history/HistoryViewer.java
```diff
@@ -0,0 +1,44 @@
...
+        setUpWindow();
+    }
+
+    private void setUpWindow() {
+        this.setMinimumSize(new Dimension(600, 300));
+        this.setTitle("History");
+        JToolBar toolBar = new JToolBar();
+        JButton clearButton = new JButton("Clear history");
+        clearButton.addActionListener((ActionEvent e) -> {
+            CommandHistory.clearHistory();
```
kowallus, line 31: You can try to remove this dependency (e.g., introducing an interface)

kowallus (changes requested):
A good starting point for further work.

src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
GGuinea: Done

PR #76, opened 2020-06-03
[IO-2-dev] small fix

adrian61:


src/main/java/edu/kis/powp/jobs2d/features/Readers/SimpleFormatReader.java
```diff
@@ -24,7 +24,7 @@ public ComplexCommand read(String contentToParse) {
             }else if(command[0].equals("operateTo")){
                 commandList.add(new OperateToCommand(Integer.valueOf(command[1]),Integer.valueOf(command[2])));
             }else{
-                System.out.println("Blad odczytu komendy w lini " + (i+1));
+                System.out.println("Error reading command on line " + (i+1));
```
kowallus, line 27: I think that "at line" would be correct

kowallus (approved):

PR #77, opened 2020-06-03
InkUsage in logs

Blackwatch98:
- still working on connecting with obeserver


src/main/java/edu/kis/powp/jobs2d/drivers/InkUsageDriver.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private int x0, y0;
+    private double totalUsage;
+    Logger logger = Logger.getLogger("global");
+    private DrawPanelController drawController;
+    private ILine line;
+
+
+
+    public InkUsageDriver(ILine line, DrawPanelController drawController){
```
kowallus, line 18: Implementing another drawing driver is not necessary.  Use decorator pattern approach.
Too many responsibilities in this class. Even the usage aspect should be extracted (e.g. separate interface)

kowallus (changes requested):
SRP violation!

PR #78, opened 2020-06-03
IO-1: Monitoring usage of a device

maybegamer:
Krzysztof Chybicki 215693
Michał Ogłoza 215834


src/main/java/edu/kis/powp/jobs2d/drivers/SelectDriverMenuOptionListener.java
```diff
@@ -17,5 +18,5 @@ public SelectDriverMenuOptionListener(Job2dDriver driver, DriverManager driverMa
     @Override
     public void actionPerformed(ActionEvent e) {
         driverManager.setCurrentDriver(driver);
-    }
+        DriverFeature.updateDriverInfo();}
```
kowallus: This line was removed in other PR, and it should not be here! Some problems in merging changes?!

src/main/java/edu/kis/powp/jobs2d/drivers/UsageMonitor.java
```diff
@@ -0,0 +1,75 @@
...
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.text.DecimalFormat;
+import java.util.logging.Logger;
+
+import static java.lang.Math.pow;
+import static java.lang.Math.sqrt;
+
+public class UsageMonitor implements Job2dDriver {
```
kowallus: This class uses a design pattern. It would be good to add its name to the class name.

src/main/java/edu/kis/powp/jobs2d/drivers/UsageMonitor.java
```diff
@@ -0,0 +1,75 @@
...
+        return operationalDistance;
+    }
+
+    /**
+     * API go get distance made by setPosition method by all drivers
+     * @return distance made by setPosition method
+     */
+    public static double getNonOperationalDistance() {
+        return operationalDistance;
+    }
```
kowallus: The design should allow many usage monitors (e.g. per different driver). First of all, it should not be static and it would be good to provide a separate interface for usage statistics.

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -96,6 +97,11 @@ private static void setupLogger(Application application) {
         application.addComponentMenuElement(Logger.class, "OFF logging", (ActionEvent e) -> logger.setLevel(Level.OFF));
     }
 
+    private static void addSubscribers(){
+        UsageMonitorObserver usageMonitorObserver = new UsageMonitorObserver();
+        DriverFeature.getDriverManager().getChangePublisher().addSubscriber(usageMonitorObserver);
+    }
```
kowallus: the name of this method should concern the function you added. Those to lines could be added directly to setupDriver.method.

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -60,7 +61,7 @@ private static void setupDrivers(Application application) {
         DrawPanelController drawerController = DrawerFeature.getDrawerController();
         Job2dDriver driver = new LineDriverAdapter(drawerController, LineFactory.getBasicLine(), "basic");
         DriverFeature.addDriver("Line Simulator", driver);
-        DriverFeature.getDriverManager().setCurrentDriver(driver);
+        //DriverFeature.getDriverManager().setCurrentDriver(driver);
```
kowallus: why this line is commented out?

kowallus (changes requested):
Some issues to resolve.

kowallus:
please close, fix and reopen PR

src/main/java/edu/kis/powp/jobs2d/drivers/Job2dDriverDecorator.java
```diff
@@ -8,18 +8,18 @@
 import static java.lang.Math.pow;
 import static java.lang.Math.sqrt;
 
-public class UsageMonitor implements Job2dDriver {
+public class Job2dDriverDecorator implements Job2dDriver {
```
kowallus, line 11: Now the name does not reflect the UsageMonitor responsibility. Try something in between.

kowallus (changes requested):
Only one issue.

kowallus:
Waiting for a better class name proposal

PR #79, opened 2020-06-03
MACRO IO2

RafalProoo:
Added possibility to draw and record macro in the same time and ability to change type of line (driver) before start of recording macro. Refactored code to composite design pattern. Used Subscriber in order to react on changes of type of line.

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/MacroDriverComposite.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/MacroDriverComposite.java
@@ -8,7 +8,7 @@ import edu.kis.powp.jobs2d.command.SetPositionCommand;
 import java.util.ArrayList;
 import java.util.List;
 
-public class MacroDriverDecorator implements Job2dDriver {
+public class MacroDriverComposite implements Job2dDriver {
     private List<DriverCommand> driverCommandList = new ArrayList<>();
     private Job2dDriver job2dDriver;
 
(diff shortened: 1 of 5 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/MacroDriverComposite.java
```diff
@@ -8,7 +8,7 @@
 import java.util.ArrayList;
 import java.util.List;
 
-public class MacroDriverDecorator implements Job2dDriver {
+public class MacroDriverComposite implements Job2dDriver {
```
kowallus, line 11: Sorry, but it is still the decorator. You should add DriverComposite class, and MacroDriver should be stripped from a decorator pattern.

src/main/java/edu/kis/powp/jobs2d/drivers/MacroDriverComposite.java
```diff
@@ -31,11 +31,13 @@ public void clearMemory() {
     @Override
     public void setPosition(int x, int y) {
         driverCommandList.add(new SetPositionCommand(x, y));
+        job2dDriver.setPosition(x, y);
```
kowallus, line 34: and now it is a decorator

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -74,22 +76,37 @@ private static void setupDrivers(Application application) {
...
+						break;
+					case "2d device simulator - special":
+						MacroFeature.getMacroDriverComposite().setCoreJob2dDriver(specialLineDriver);
+						break;
+					case "2d device simulator - dotted":
+						MacroFeature.getMacroDriverComposite().setCoreJob2dDriver(dottedLineDriver);
+						break;
+				}
+			}
+		});
```
kowallus, line 109: Such large changes in this class are not allowed. Extract such configuration routines after fixing the composite design.

kowallus (changes requested):
Sorry, renaming the class does not change the design pattern used.

PR #80, opened 2020-06-04
IO-3-dev Importing command from text

BartoszKowalczyk98:


BartoszKowalczyk98:
Przykładowa zawartość pliku Json:
{
  "commandName": "jakis eseflores",
  "operations": [
    {
      "opName": "SetPositionCommand",
      "x": -20,
      "y": -50
    },
    {
      "opName": "OperateToCommand",
      "x": -20,
      "y": -50
    },
    {
      "opName": "OperateToCommand",
      "x": 20,
      "y": 50
    },
    {
      "opName": "OperateToCommand",
      "x": 25,
      "y": 50
    }
  ]
}

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -83,6 +102,25 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
 		content.add(btnRunCommand, c);
 	}
 
+	private void importCommands() {
+		int returnValue = fileChooser.showOpenDialog(null);
+		if (returnValue == JFileChooser.APPROVE_OPTION) {
+			File selectedFile = fileChooser.getSelectedFile();
+			JsonParser jsonParser = new JsonParser(selectedFile);
```
kowallus: The goal is to make the design of parser reusable. Now we accept the content only from a File (what if we have a simple String? we are forced to save it to File to use parser 👎 ) and we cannot add support for other formats easily.
Moreover, we cannot reuse parser instance (we need a new one for a single parse). This all should be fixed by defining a correct interface.

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -83,6 +102,25 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
+		}
+	}
+
+	private void exportCommands() {
+		fileChooser.setDialogTitle("Specify a file to save");
+		int userSelection = fileChooser.showSaveDialog(this);
+		if (userSelection == JFileChooser.APPROVE_OPTION) {
+			File fileToSave = fileChooser.getSelectedFile();
+			JsonParser jsonParser = new JsonParser(fileToSave);
+			jsonParser.parseToExport(commandManager.getCurrentCommand());
```
kowallus: The same problem

src/main/java/edu/kis/powp/jobs2d/command/manager/parsing/JsonParser.java
```diff
@@ -0,0 +1,107 @@
...
+import edu.kis.powp.jobs2d.command.OperateToCommand;
+import edu.kis.powp.jobs2d.command.SetPositionCommand;
+
+import java.io.*;
+import java.util.ArrayList;
+import java.util.List;
+
+public class JsonParser implements Parser {
+	File jsonFile;
+	String commandName = "No command has been set yet";
```
kowallus: These should be private (however, I do not like a dependency to a File)

src/main/java/edu/kis/powp/jobs2d/command/manager/parsing/JsonParser.java
```diff
@@ -0,0 +1,107 @@
...
+			JsonDriverCommand jsonDriverCommand = gson.fromJson(reader, JsonDriverCommand.class);
+			List<DriverCommand> driverCommands = new ArrayList<>();
+			for (JsonDriverCommand.operation op : jsonDriverCommand.operations)
+				if (op.opName.equals("SetPositionCommand")) driverCommands.add(new SetPositionCommand(op.x, op.y));
+				else if (op.opName.equals("OperateToCommand")) driverCommands.add(new OperateToCommand(op.x, op.y));
+			commandName = jsonDriverCommand.commandName;
+			return driverCommands;
+		} catch (FileNotFoundException e) {
+			e.printStackTrace();
+		}
```
kowallus: weak exception handling

src/main/java/edu/kis/powp/jobs2d/command/manager/parsing/Parser.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.io.File;
+import java.util.List;
+
+public interface Parser {
+	List<DriverCommand> parseFromImport();
+
+	String getCommandName();
+
+	File parseToExport(DriverCommand driverCommand);
```
kowallus: Again, why File?

src/main/java/edu/kis/powp/jobs2d/command/manager/parsing/Parser.java
```diff
@@ -0,0 +1,14 @@
+package edu.kis.powp.jobs2d.command.manager.parsing;
+
+import edu.kis.powp.jobs2d.command.DriverCommand;
+
+import java.io.File;
+import java.util.List;
+
+public interface Parser {
+	List<DriverCommand> parseFromImport();
```
kowallus: The interface is inconsistent!!! It defines an export format, but no import format. Internal format for import is List of commands and for export, it is a DriverCommand.

src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverDecorator.java
```diff
@@ -40,11 +40,13 @@ public Job2dDriver getDriver() {
 	@Override
 	public void setPosition(int x, int y) {
 		driverCommandList.add(new SetPositionCommand(x, y));
+		driver.operateTo(x,y);
```
kowallus: What happened here?! The lines added here are not related to your task? Where did you get those (copy&paste?)?!

kowallus (changes requested):
Some issues, suggestions and questions?

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
AndrzejBilant: At "we cannot add support for other formats easily". There is a filter method in FileChooser and all we have to do is add format after a comma.

src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverDecorator.java
BartoszKowalczyk98: This was to solve that "Bug: drawing does not work during recording MACRO" and it slipped with one commit and i didn't notice it go now but it is now drawing lines during recording macro. So that's what happend.

kowallus: OK, but it would be nice to have it in a separate PR (but this require working on different branches on different features)

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -58,6 +69,43 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
+			} catch (IOException exception) {
+				exception.printStackTrace();
+			}
+		});
+		c.fill = GridBagConstraints.BOTH;
+		c.weightx = 1;
+		c.gridx = 0;
+		c.weighty = 1;
+		content.add(btnExportCommand, c);
+
```
kowallus: This method is large now. It would be best to decompose it to several smaller.

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -83,6 +131,24 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
 		content.add(btnRunCommand, c);
 	}
 
+	private void importCommands() {
+		parser = new JsonParser();
```
kowallus: This parser could be a local variable (it is used this way). It would be better if the JsonParser object could be reused (then it can be a field).

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -83,6 +131,24 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
+		commandManager.setCurrentCommand(parser.parseFromString(txtFieldToImport.getText()));
+	}
+
+	private void exportCommands() throws IOException {
+		fileChooser.setDialogTitle("Specify a file to save");
+		int userSelection = fileChooser.showSaveDialog(this);
+		if (userSelection == JFileChooser.APPROVE_OPTION) {
+			File fileToSave = fileChooser.getSelectedFile();
+			BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(fileToSave));
+			parser = new JsonParser();
```
kowallus: the same issue.

src/main/java/edu/kis/powp/jobs2d/command/manager/parsing/JsonParser.java
```diff
@@ -0,0 +1,79 @@
...
+		Gson gson = new Gson();
+		JsonDriverCommand jsonDriverCommand = new JsonDriverCommand();
+
+		jsonDriverCommand.operations = getOperationFromDriverCommand(driverCommand);
+		return gson.toJson(jsonDriverCommand, JsonDriverCommand.class);
+	}
+
+	private List<JsonDriverCommand.operation> getOperationFromDriverCommand(DriverCommand driverCommand) {
+		List<JsonDriverCommand.operation> operationList = new ArrayList<>();
+		if (driverCommand instanceof OperateToCommand) {
```
kowallus, line 51: Only an optional suggestion: instanceof is not object-oriented approach. If you want to make it right you can use, e.g. visitor pattern.

kowallus (changes requested):
One mandatory code cleaning suggestion.

kowallus:
You can work on taking advantage of the visitor pattern to fix unnecessary instanceof as your second task.

PR #81, opened 2020-06-04
Io 2 dev History

GGuinea:
215851
215793
215671
215933
+ NPE fixed


src/main/java/edu/kis/powp/jobs2d/command/ImmutableComplexCommand.java
```diff
@@ -8,10 +8,13 @@
 import edu.kis.powp.jobs2d.Job2dDriver;
 
 public class ImmutableComplexCommand implements ICompoundCommand {
+
     private List<DriverCommand> driverCommandList;
+    private String commandName;
 
-    public ImmutableComplexCommand(List<DriverCommand> driverCommandList) {
+    public ImmutableComplexCommand(List<DriverCommand> driverCommandList, String commandName) {
```
kowallus, line 15: It would be safer to add (not modify) the constructor. The old one might be in use by other (work in progress) features. You can add a default commandName (e.g. "unknown immutable command")

src/main/java/edu/kis/powp/jobs2d/command/history/HistoryEntry.java
```diff
@@ -0,0 +1,29 @@
...
+    private DriverCommand command;
+
+    public String getCommandName() {
+        return commandName;
+    }
+
+    public HistoryEntry(String date, String commandName, DriverCommand command) {
+        this.date = date;
+        this.commandName = commandName;
+        this.command = command;
```
kowallus, line 18: Can you ensure that the command passed by the parameter won't be changed by others?

src/main/java/edu/kis/powp/jobs2d/command/history/HistoryViewer.java
```diff
@@ -0,0 +1,40 @@
...
+    private DriverCommandManager driverCommandManager;
+
+    public HistoryViewer(DefaultListModel<HistoryEntry> historyCommandList, DriverCommandManager driverCommandManager)
+            throws HeadlessException {
+        this.historyCommandList = new JList<HistoryEntry>(historyCommandList);
+        this.driverCommandManager = driverCommandManager;
+        setUpWindow();
+    }
+
+    private void setUpWindow() {
```
kowallus: setupWindow

src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
```diff
@@ -0,0 +1,27 @@
...
+import java.time.LocalTime;
+import java.time.format.DateTimeFormatter;
+
+import javax.swing.DefaultListModel;
+
+import edu.kis.powp.jobs2d.command.DriverCommand;
+
+public class CommandHistory {
+
+    private static DefaultListModel<HistoryEntry> entryHistoryList = new DefaultListModel<>();
```
kowallus, line 12: a declaration should be limited to the interface, i.e. ListModel.

src/main/java/edu/kis/powp/jobs2d/events/SelectHistoryListOptionListener.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.swing.event.ListSelectionEvent;
+import javax.swing.event.ListSelectionListener;
+
+import edu.kis.powp.jobs2d.command.history.HistoryEntry;
+import edu.kis.powp.jobs2d.command.manager.DriverCommandManager;
+
+public class SelectHistoryListOptionListener implements ListSelectionListener {
+
+    private DriverCommandManager commandManager;
+    private final JList<HistoryEntry> historyCommandList;
```
kowallus: I think that the listener could be independent of JList implementation but rather relying only on ListModel.

src/main/java/edu/kis/powp/jobs2d/events/SelectHistoryListOptionListener.java
```diff
@@ -0,0 +1,31 @@
...
+    public SelectHistoryListOptionListener(DriverCommandManager commandManager,
+            JList<HistoryEntry> historyCommandList) {
+        this.commandManager = commandManager;
+        this.historyCommandList = historyCommandList;
+    }
+
+    @Override
+    public void valueChanged(ListSelectionEvent e) {
+        if (e.getValueIsAdjusting() == false) {
+            HistoryEntry selectedValue = historyCommandList.getSelectedValue();
```
kowallus: I think that 'e' holds info about an index of selected command which could be used to obtain a HistoryEntry from listModel. Please check this.

kowallus (changes requested):
Mostly minor issues, but I would like you to check the possibility to make history facilities more independent of GUI (by using a shared  model)

kowallus:
Please fix minor issues first, so I can merge the code.

kowallus:
waiting...

src/main/java/edu/kis/powp/jobs2d/command/history/HistoryEntry.java
GGuinea: Yes, in DriverCommandManager in method setCurrentCommand we are creating new ImmutableComplexCommand which later is going to be present in history.

src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
GGuinea: It can not be like that. ListModel does not contains addElement method, which we are using.

src/main/java/edu/kis/powp/jobs2d/events/SelectHistoryListOptionListener.java
GGuinea: We have chosen JList because of the ability of selecting elements from list using mouse.

src/main/java/edu/kis/powp/jobs2d/events/SelectHistoryListOptionListener.java
GGuinea: "e" does not provides easy access to chosen element.

src/main/java/edu/kis/powp/jobs2d/command/ImmutableComplexCommand.java
GGuinea: Done

src/main/java/edu/kis/powp/jobs2d/command/history/HistoryViewer.java
GGuinea: Done

src/main/java/edu/kis/powp/jobs2d/command/history/CommandHistory.java
kowallus: ok

src/main/java/edu/kis/powp/jobs2d/command/history/HistoryEntry.java
kowallus: ok

src/main/java/edu/kis/powp/jobs2d/events/SelectHistoryListOptionListener.java
kowallus: Yes, but internally JList relies on ListModel as a data holder.

src/main/java/edu/kis/powp/jobs2d/events/SelectHistoryListOptionListener.java
kowallus: e doesn't, but I think that a ListModel object does. It would be good to decouple from GUI

kowallus (changes requested):
still I have some doubts...

src/main/java/edu/kis/powp/jobs2d/events/SelectHistoryListOptionListener.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public SelectHistoryListOptionListener(DriverCommandManager commandManager, DefaultListModel<HistoryEntry> historyCommandList) {
+        this.commandManager = commandManager;
+        this.historyCommandList = historyCommandList;
+    }
+
+    @Override
+    public void valueChanged(ListSelectionEvent e) {
+        if (e.getValueIsAdjusting() == false) {
+            final int selectedIndex = ((JList<HistoryEntry>) e.getSource()).getSelectedIndex();
```
kowallus, line 23: Casting JList is a worse variant. My tip concerning DefaultListModel and ListSelectionEvent was probably incorrect.
Nevertheless, it is possible to avoid the direct dependency with JList, by introducing a new abstraction (an interface) which will encapsulate retrieving a selected index.

kowallus (approved):
Only a comment

PR #82, opened 2020-06-05
IO4 - Usable commands factory

startled-cat:


Luvkitri:
Seems about right

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -97,7 +98,35 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
+			commandManager.commandFactory.clear();
+			cfChoice.removeAll();
+		});
+		btnFactorySetCommand.addActionListener(e -> {
+			commandManager.setCurrentCommand(cfChoice.getSelectedItem());
+		});
+
+		content.add(btnFactoryClearCommand, c);
+		content.add(btnFactorySetCommand, c);
+
```
kowallus: Extract factory gui configuration to a separate method.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -37,8 +41,27 @@ public DriverCommandManager(DriverManager driverManager) {
 	 * @param name        name of the command.
 	 */
 	public synchronized void setCurrentCommand(List<DriverCommand> commandList, String name) {
+		commandFactory.add(name, new DefaultCompoundCommand((commandList)));
+		currentCommandName = name;
```
kowallus: OK. Actually 'name' should be added to the DefaultCompoundCommand (and it should be returned by toString)

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -37,8 +41,27 @@ public DriverCommandManager(DriverManager driverManager) {
 	 * @param name        name of the command.
 	 */
 	public synchronized void setCurrentCommand(List<DriverCommand> commandList, String name) {
+		commandFactory.add(name, new DefaultCompoundCommand((commandList)));
```
kowallus: No. You should not add the dependency to the commandFactory here. Use observer (see publisher)! This line should be implemented in a separate commandManager subscriber implementation.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -37,8 +41,27 @@ public DriverCommandManager(DriverManager driverManager) {
...
 		setCurrentCommand(new DefaultCompoundCommand(commandList));
 	}
+	/**
+	 * Sets current command.
+	 *
+	 * @param name        name of the command.
+	 */
+	public synchronized void setCurrentCommand(String name) {
+		setCurrentCommand(commandFactory.get(name));
+	}
```
kowallus: This also should be removed. You can call setCurrentCommand from the factory (without creating reverse dependency)

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import edu.kis.powp.jobs2d.command.ImmutableCompoundCommand;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Set;
+
+public class CommandFactory {
+
+    private HashMap<String, ICompoundCommand> shapes;
```
kowallus: The declaration should indicate only the interface, i.e. Map.
I think that the type of values should also be more general, i.e., DriverCommand

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,45 @@
...
+        shapes = new HashMap<>();
+    }
+
+    public ICompoundCommand get(String name){
+        return shapes.get(name);
+    }
+
+    public void add(String name, ICompoundCommand commands) throws IllegalArgumentException{
+        try{
+            shapes.put(name, (DefaultCompoundCommand) commands.clone());
```
kowallus: Is casting to DefaultCompoundCommand really important?

src/test/java/edu/kis/powp/jobs2d/features/CommandFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+        cf.get("line").forEach(listFromfactory::add);
+
+        assertEquals(listOriginal.size(), listFromfactory.size());
+
+        for (int i = 0; i < listOriginal.size(); i++) {
+            if(listOriginal.get(i) instanceof OperateToCommand){
+                assert(listFromfactory.get(i) instanceof OperateToCommand);
+                OperateToCommand a1 = (OperateToCommand)listOriginal.get(i);
+                OperateToCommand a2 = (OperateToCommand)listOriginal.get(i);
+                assertEquals(a1.getX(), a2.getX());
```
kowallus, line 89: How could possibly a1 differ from a2 if they are both listOriginal.get(i)?! This part of the test does not make sense.

src/test/java/edu/kis/powp/jobs2d/features/CommandFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+    @Test
+    void getNamesOfStored() {
+        CommandFactory cf = new CommandFactory();
+
+        List<DriverCommand> dcList = new ArrayList<>();
+        dcList.add(new SetPositionCommand(100, 100));
+        dcList.add(new OperateToCommand(150, 150));
+        ImmutableCompoundCommand icc = new ImmutableCompoundCommand(dcList);
+
+        cf.add("line", icc);
```
kowallus: You should better use constants.

kowallus (changes requested):
Some minor issues and to many changes in DriverCommandManager makes merging impossible.

kowallus:
please close, fix and reopen PR

PR #83, opened 2020-06-05
IO-1 Visitor and Commands transformation

Miretpl:
Patryk Janowski
Przemysław Mirowski

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/transformation/CommandTransformation.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/transformation/CommandTransformation.java
@@ -1,0 +1,45 @@
+package edu.kis.powp.jobs2d.command.transformation;
+
+import edu.kis.powp.jobs2d.command.*;
+import edu.kis.powp.jobs2d.command.visitor.CommandVisitorInterface;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class CommandTransformation implements CommandVisitorInterface {
+    private final List<DriverCommand> driverCommandList;
+
+    public CommandTransformation() {
+        driverCommandList = new ArrayList<>();
+    }
+
+    public ICompoundCommand getCompoundCommand() {
+        return new DefaultCompoundCommand(driverCommandList);
+    }
+
+    @Override
+    public void visit(SetPositionCommand setPositionCommand) {
+        int transformedXPoint = transformXPoint(setPositionCommand.getPosX(), setPositionCommand.getPosY());
+        int transformedYPoint = transformYPoint(setPositionCommand.getPosX(), setPositionCommand.getPosY());
+
+        driverCommandList.add(new SetPositionCommand(transformedXPoint, transformedYPoint));
+    }
+
+    @Override
+    public void visit(OperateToCommand operateToCommand) {
+        int transformedXPoint = transformXPoint(operateToCommand.getPosX(), operateToCommand.getPosY());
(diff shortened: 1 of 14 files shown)


src/test/java/edu/kis/powp/jobs2d/events/SelectCommandTransformationRotateOptionListener.java
```diff
@@ -0,0 +1,30 @@
...
+import edu.kis.powp.jobs2d.command.transformation.CommandTransformation;
+import edu.kis.powp.jobs2d.command.transformation.CommandTransformationRotate;
+import edu.kis.powp.jobs2d.features.CommandsFeature;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+public class SelectCommandTransformationRotateOptionListener implements ActionListener {
+
+    double angle;
```
kowallus, line 14: should be private

src/test/java/edu/kis/powp/jobs2d/events/SelectCommandTransformationFlipOptionListener.java
```diff
@@ -0,0 +1,29 @@
...
+import edu.kis.powp.jobs2d.command.manager.DriverCommandManager;
+import edu.kis.powp.jobs2d.command.transformation.CommandTransformation;
+import edu.kis.powp.jobs2d.command.transformation.CommandTransformationFlip;
+import edu.kis.powp.jobs2d.features.CommandsFeature;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+
+public class SelectCommandTransformationFlipOptionListener implements ActionListener {
+    boolean isHorizontal;
```
kowallus, line 13: should be private

src/main/java/edu/kis/powp/jobs2d/command/transformation/CommandTransformation.java
```diff
@@ -0,0 +1,45 @@
+package edu.kis.powp.jobs2d.command.transformation;
+
+import edu.kis.powp.jobs2d.command.*;
+import edu.kis.powp.jobs2d.command.visitor.CommandVisitorInterface;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class CommandTransformation implements CommandVisitorInterface {
```
kowallus, line 9: This class is used according to visitor pattern so the name should state it explicitly. Maybe we could avoid exposing a visitor interface to clients?

src/main/java/edu/kis/powp/jobs2d/command/transformation/CommandTransformation.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    public void visit(ICompoundCommand iCompoundCommand){
+        for (DriverCommand driverCommand : iCompoundCommand) {
+            driverCommand.accept(this);
+        }
+    }
+
+    abstract int transformXPoint(int x, int y);
+    abstract int transformYPoint(int x, int y);
```
kowallus, line 44: Optional future improvement: 
Those methods are not strongly consistent with the rest of this class (they do not operate on any fields of CommandTransformation). Why not making this aspect an independent re-usable hierarchy (e.g. strategy or bridge pattern)? This probably should also make core and testing code simpler.

kowallus (approved):
Well done. However, I have still one suggestion for a better final grade.

PR #84, opened 2020-06-06
IO-1 Visitor and Commands transformation refactor

Miretpl:
Patryk Janowski
Przemysław Mirowski

Related to #83


src/test/java/edu/kis/powp/jobs2d/events/SelectCommandTransformationScaleOptionListener.java
```diff
@@ -2,16 +2,16 @@
...
 import edu.kis.powp.jobs2d.features.CommandsFeature;
 
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 
 public class SelectCommandTransformationScaleOptionListener implements ActionListener {
-    double scaleFactorX;
-    double scaleFactorY;
+    private final double scaleFactorX;
+    private final double scaleFactorY;
```
kowallus, line 15: Could all command transformation listeners could be implemented as one listener (injecting trasformation instead of particular transformation params)

kowallus (approved):
OK. Only one observation.

PR #85, opened 2020-06-06
MACRO IO2

RafalProoo:


src/main/java/edu/kis/powp/jobs2d/drivers/DriverComposite.java
```diff
@@ -0,0 +1,51 @@
...
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+import edu.kis.powp.jobs2d.command.DriverCommand;
+import edu.kis.powp.jobs2d.command.Visitor;
+import edu.kis.powp.jobs2d.features.DriverFeature;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class DriverComposite implements DriverCommand {
```
kowallus, line 11: Sorry, but this is not a DriverComposite (composite in Job2dDriver hierarchy) but DriverCommandComposite (which is not a topic of this task).

kowallus (changes requested):
This is not right. Reset and start driver composite again!

PR #87, opened 2020-06-08
Visitor design pattern in commands

bopke:


kowallus:
Have you checked project after 'Resolve Conflicts' commit? It looks like there should be errors in your branch now.

PR #88, opened 2020-06-08
IO-4 View Transformation

Kristrox:


src/main/java/edu/kis/powp/jobs2d/drivers/Jobs2dDriverTransformationDecorator.java
```diff
@@ -0,0 +1,24 @@
...
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class Jobs2dDriverTransformationDecorator implements Job2dDriver{
+    protected Job2dDriver driver;
+    private int xTrans=1;
+    private int yTrans=1;
+    public Jobs2dDriverTransformationDecorator(Job2dDriver driver, int xTrans, int yTrans) {
+        this.driver = driver;
+        this.xTrans *= xTrans;
+        this.yTrans *= yTrans;
```
kowallus, line 12: What is the point of using *= instead of =?

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -84,26 +86,32 @@ private static void setupDrivers(Application application) {
...
+
 		DriverFeature.addDriver("Logger driver", loggerDriver);
-		
+
 		DrawPanelController drawerController = DrawerFeature.getDrawerController();
 		Job2dDriver driver = new LineDriverAdapter(drawerController, LineFactory.getBasicLine(), "basic");
 		DriverFeature.addDriver("Line Simulator", driver);
 		DriverFeature.getDriverManager().setCurrentDriver(driver);
+
+		DriverFeature.addDriver("Scale x2", new Jobs2dDriverTransformationDecorator(DriverFeature.getDriverManager().getCurrentDriver(), 2, 2));
```
kowallus, line 100: Transformation can work with different drivers but it is fixed. You should use driver change observer to transparently apply transformations to various drivers.

src/main/java/edu/kis/powp/jobs2d/drivers/Jobs2dDriverTransformationDecorator.java
```diff
@@ -0,0 +1,24 @@
+package edu.kis.powp.jobs2d.drivers;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+public class Jobs2dDriverTransformationDecorator implements Job2dDriver{
```
kowallus, line 5: The design should enable support for a wider range of transformations, e.g. movement, rotation, and enable their compositions.

kowallus (changes requested):
Nice but the solution should be more extendible.

PR #90, opened 2020-06-11
Io 3 dev history of commands

MR-98:


src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -81,6 +83,24 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
 		c.gridx = 0;
 		c.weighty = 1;
 		content.add(btnRunCommand, c);
+
+		DefaultListModel model = new DefaultListModel();
```
kowallus, line 87: Declarations should focus on the interface (ListModel)

src/main/java/edu/kis/powp/jobs2d/command/manager/CommandHistoryController.java
```diff
@@ -0,0 +1,39 @@
...
+
+import edu.kis.powp.jobs2d.command.DriverCommand;
+
+import javax.swing.*;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CommandHistoryController {
+
+    static DefaultListModel listModel;
```
kowallus, line 11: should be private and focused on interface (ListModel)

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -46,6 +48,9 @@ public synchronized void runCurrentCommand() {
 	 */
 
 	public synchronized void setCurrentCommand(List<DriverCommand> commandList, String name) {
+		this.commandList = commandList;
+		this.commandName = name;
+
```
kowallus, line 53: Do not introduce redundant CommandList (because history should work with DriverCommands, not list of commands)
Instead commandName, you should add name to CompoundCommand.

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -72,4 +77,12 @@ public synchronized String getCurrentCommandString() {
...
 		return changePublisher;
 	}
+
+	public synchronized List<DriverCommand> getCommandList(){
+		return commandList;
+	}
+
+	public synchronized String getCommandName(){
+		return commandName;
+	}
```
kowallus, line 87: Should be removed.

src/main/java/edu/kis/powp/jobs2d/command/manager/CommandHistoryController.java
```diff
@@ -0,0 +1,39 @@
...
+    static List<CommandHistoryEntry> historyEntries = new ArrayList<>();
+
+    public static void addNewHistoryEntry(List<DriverCommand> commands, String name){
```
kowallus, line 14: An entry should be a DriverCommand not a list of commands.

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -90,6 +110,7 @@ private void clearCommand() {
 
 	private void runCommand() {
 		commandManager.runCurrentCommand();
+		CommandHistoryController.addNewHistoryEntry(CommandsFeature.getDriverCommandManager().getCommandList(),CommandsFeature.getDriverCommandManager().getCommandName());
```
kowallus, line 113: commandManager.runCurrentCommand() can be executed from other places. addNewHistoryEntry should be triggered by this command.

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -81,6 +83,24 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
+			if (!e.getValueIsAdjusting()) {
+
+				if (commandHistoryList.getSelectedIndex() != -1) {
+					int index = commandHistoryList.getSelectedIndex();
+					DriverCommandManager manager = CommandsFeature.getDriverCommandManager();
+
+					manager.setCurrentCommand(CommandHistoryController.getCommandsFromList(index),CommandHistoryController.getCommandsNameFromList(index));
+				}
+			}
+		});
```
kowallus, line 102: Move lambda to a separate class. Maybe some dependencies can be avoided?

kowallus (changes requested):
Please address the following issues

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
MR-98: ListModel doesn't have add() method which we are using, have to stay

PR #93, opened 2020-06-12
Usage Monitor IO-3

Smirnoffq:


src/main/java/edu/kis/powp/jobs2d/features/UsageMonitor.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Iterator;
+import java.util.logging.Logger;
+
+public class UsageMonitor implements CommandVisitorInterface {
+
+    private Job2dDriver job2dDriver;
+    private int operateToDistance;
+    private int positionDistance;
+    private int currentX;
+    private int currentY;
```
kowallus: Attributes should be explicitly initialized.

src/test/java/edu/kis/powp/jobs2d/UsageMonitorTest.java
```diff
@@ -0,0 +1,74 @@
...
+        driverCommands.add(new OperateToCommand(-20, 50));
+        driverCommands.add(new SetPositionCommand(0, -50));
+        driverCommands.add(new SetPositionCommand(-20, -50));
+        driverCommands.add(new OperateToCommand(-20, -50));
+        driverCommands.add(new SetPositionCommand(-20, -40));
+        driverCommands.add(new OperateToCommand(-20, 50));
+        driverCommands.add(new SetPositionCommand(0, -50));
+        driverCommands.add(new OperateToCommand(0, -50));
+        driverCommands.add(new SetPositionCommand(0, -40));
+        driverCommands.add(new OperateToCommand(0, 50));
```
kowallus: You could extract it to separate commands factory, add load option (setCurrentCommand) to GUI and use getCurrentCommand to apply visitor to any command.

src/test/java/edu/kis/powp/jobs2d/UsageMonitorTest.java
```diff
@@ -0,0 +1,74 @@
...
+            public Iterator<DriverCommand> iterator()
+            {
+                return commands.iterator();
+            }
+
+            @Override
+            public void execute(Job2dDriver driver)
+            {
+                commands.forEach(c -> c.execute(driver));
+            }
```
kowallus: Ain't there implementations of ICompoundCommand available?

src/main/java/edu/kis/powp/jobs2d/features/UsageMonitor.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public void visit(OperateToCommand driver)
+    {
+        operateToDistance += calculateDistance(driver.getPosX(), driver.getPosY());
+        currentX = driver.getPosX();
+        currentY = driver.getPosY();
+
+        logger.info("Operate distance: " + operateToDistance + "\nHead distance: " + positionDistance);
```
kowallus: instead of constant logging, it would be better to implement getters and use them to log after visitor job.

kowallus (changes requested):
You have implemented the wrong task. Instead of "Monitoring usage of a device" (working on a driver) you created a command usage evaluator.

kowallus:
please close, update and reopen

PR #94, opened 2020-06-12
Io 1 dev

PrzemaXster:
Grupa:
Przemysław Grabarek 215731
Tomasz Majcher 215810
zadanie: MACRO - recording driver calls as commands
Dodano funckję load Macro command do TestJobs2dApp

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/ChangeTitleObserver.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/ChangeTitleObserver.java
@@ -1,0 +1,16 @@
+package edu.kis.powp.jobs2d.drivers;
+
+import edu.kis.powp.jobs2d.features.DriverFeature;
+import edu.kis.powp.observer.Subscriber;
+
+
+public class ChangeTitleObserver implements Subscriber {
+
+    public void update() {
+        DriverFeature.updateDriverInfo();
+    }
+
+    public String toString() {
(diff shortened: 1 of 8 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/ChangeTitleObserver.java
```diff
@@ -0,0 +1,16 @@
...
+public class ChangeTitleObserver implements Subscriber {
+
+    public void update() {
+        DriverFeature.updateDriverInfo();
+    }
+
+    public String toString() {
+        return "Driver Title Bar Observer";
+    }
+}
```
kowallus, line 16: 1. This is not related with your task (and all other changes associated with observer pattern). How is it possible that you have implemented it?
2. It has already been implemented in #67. Your repo is seriously outdated: "This branch is 3 commits ahead, 67 commits behind iis-powp-2020:IO-1-dev."! You must merge the changes from the main branch.

kowallus (changes requested):
I am sorry to see such a plagiarism attempt. 
I will not count this task to your grade, but if you want to pass the subject, you need to fix the flows in the design and in the work of this feature. First of all, you need to remove irrelevant modifications and update your repo. 
Later you can work on improving the feature: the driver selection should be independent of the macro option.
After this, I will think about another task for you to pass the project.

PR #96, opened 2020-06-14
IO-3 Bug: drawing does not work during recording MACRO

BartoszKowalczyk98:
[bugfix] drawing during Macro recording
Signed-off-by: Kowalczyk.Bartosz <215772@edu.p.lodz.pl>

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverDecorator.java
+++ b/src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverDecorator.java
@@ -40,11 +40,13 @@ public class MacroDriverDecorator implements Job2dDriver {
 	@Override
 	public void setPosition(int x, int y) {
 		driverCommandList.add(new SetPositionCommand(x, y));
+		driver.setPosition(x,y);
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverDecorator.java
```diff
@@ -7,48 +7,60 @@
 
 import java.util.ArrayList;
 import java.util.List;
-
+/**Driver decorator that records executed commands*/
```
kowallus, line 10: Not really the javadoc convention

src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverDecorator.java
```diff
@@ -7,48 +7,60 @@
...
 public class MacroDriverDecorator implements Job2dDriver {
+	/**List of operations*/
 	private List<DriverCommand> driverCommandList;
-
+	/**Driver*/
```
kowallus, line 14: Really???

kowallus (approved):
Commenting could be better (and sometimes less is better)

PR #104, opened 2020-06-16
MACRO IO2

RafalProoo:


src/main/java/edu/kis/powp/jobs2d/drivers/DriverComposite.java
```diff
@@ -0,0 +1,41 @@
+package edu.kis.powp.jobs2d.drivers;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class DriverComposite implements Job2dDriver {
+
+    List<Job2dDriver> job2dDriverList;
```
kowallus: should be private

src/main/java/edu/kis/powp/jobs2d/drivers/MacroDriver.java
```diff
@@ -8,34 +8,36 @@
 import java.util.ArrayList;
 import java.util.List;
 
-public class MacroDriverDecorator implements Job2dDriver {
+public class MacroDriver implements Job2dDriver {
     private List<DriverCommand> driverCommandList = new ArrayList<>();
-    private Job2dDriver job2dDriver;
+    private DriverComposite driverComposite;
```
kowallus: No, macro driver should not contain a driver composite! Macro should only perform macro recording funcion (SRP principle)

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -72,22 +70,26 @@ private static void setupDrivers(Application application) {
...
 
-		driver = new LineDriverAdapter(drawerController, LineFactory.getDottedLine(), "dotted");
-		DriverFeature.addDriver("Dotted line Simulator", driver);
+		DriverFeature.addDriver("Start Macro Driver", MacroFeature.getMacroDriver());
 
-		DriverFeature.addDriver("Start Macro Driver", MacroFeature.getMacroDriverDecorator());
-		MacroFeature.getMacroDriverDecorator().setCoreJob2dDriver(driver);
+		DriverComposite driverComposite = new DriverComposite();
+		driverComposite.addDriver(basicLineDriver);
+		MacroFeature.getMacroDriver().setDriverComposite(driverComposite);
```
kowallus: No like this! The other way around:
driverComposite.addDriver(MacroFeature.getMacroDriver());
Is it clear now? ;)

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -72,22 +70,26 @@ private static void setupDrivers(Application application) {
...
     
-    DriverInfoChangeObserver driverInfoChangeObserver = new DriverInfoChangeObserver();
+    	DriverInfoChangeObserver driverInfoChangeObserver = new DriverInfoChangeObserver();
```
kowallus: Original code convention used tabs (not spaces)

kowallus (changes requested):
Inconsistency in using a composite pattern. The macro driver should now be simplified.

src/main/java/edu/kis/powp/jobs2d/features/DriverChangeObserver.java
```diff
@@ -0,0 +1,29 @@
...
+import edu.kis.powp.jobs2d.drivers.adapter.LineDriverAdapter;
+import edu.kis.powp.observer.Subscriber;
+
+public class DriverChangeObserver implements Subscriber {
+
+    @Override
+    public void update() {
+        Job2dDriver currentDriver = DriverFeature.getDriverManager().getCurrentDriver();
+
+        if (currentDriver instanceof MacroDriver) {
```
kowallus, line 17: instanceof is a code smell for a bad OO design. Actually I do not see any reason for this Subscriber (DriverComposite with Macro feature could be simply added to drivers). 
If Macro could be added to any driver then Subscriber would be needed.

kowallus (approved):
Driver composite design is correct. However, the goal described in trello task features has not been reached.

PR #105, opened 2020-06-16
Add visitor pattern tests

bopke:
As determined a few days ago

### DIFF
--- a/src/test/java/edu/kis/powp/jobs2d/CommandUsageCounterVisitor.java
+++ b/src/test/java/edu/kis/powp/jobs2d/CommandUsageCounterVisitor.java
@@ -4,8 +4,8 @@ import edu.kis.powp.jobs2d.command.*;
 import edu.kis.powp.jobs2d.command.visitor.CommandVisitorInterface;
 
 public class CommandUsageCounterVisitor implements CommandVisitorInterface {
-    public int operateToCounter = 0;
-    public int setPositionCounter = 0;
+    private int operateToCounter = 0;
+    private int setPositionCounter = 0;
 
     public void visit(ICompoundCommand command) {
         for (DriverCommand partOfCompoundCommand : command) {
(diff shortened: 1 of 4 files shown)


src/test/java/edu/kis/powp/jobs2d/CommandUsageCounterVisitor.java
```diff
@@ -0,0 +1,24 @@
+package edu.kis.powp.jobs2d;
+
+import edu.kis.powp.jobs2d.command.*;
+import edu.kis.powp.jobs2d.command.visitor.CommandVisitorInterface;
+
+public class CommandUsageCounterVisitor implements CommandVisitorInterface {
+    public int operateToCounter = 0;
+    public int setPositionCounter = 0;
```
kowallus: These fields should be private and accessed only using getters (so counting wouldn't be corrupted from outside)

src/test/java/edu/kis/powp/jobs2d/events/SelectVisitorTestListener.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.awt.event.ActionEvent;
+import java.awt.event.ActionListener;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+import java.util.logging.Logger;
+
+public class SelectVisitorTestListener implements ActionListener {
```
kowallus: Name should explain the function: e.g., SelectCommandUsageCounterVisitorTestListener.

src/test/java/edu/kis/powp/jobs2d/events/SelectVisitorTestListener.java
```diff
@@ -0,0 +1,78 @@
...
+            @Override
+            public void execute(Job2dDriver driver) {
+
+            }
+
+            @Override
+            public DriverCommand clone() throws CloneNotSupportedException {
+                return null;
+            }
+        };
```
kowallus: Such command should be used for unit tests.
Inside application counting should be applied for a user configurated command. Please simply use a command available through driverCommandManager.getCurrentCommand()

kowallus (changes requested):
Few simple issues to fix

PR #106, opened 2020-06-16
Immutable complex command

michaaeu:


src/main/java/edu/kis/powp/jobs2d/command/ImmutableComplexCommand.java
```diff
@@ -10,8 +10,16 @@
...
 
-    public ImmutableComplexCommand(List<DriverCommand> driverCommandList) {
-        this.driverCommandList = Collections.unmodifiableList(driverCommandList);
+    public ImmutableComplexCommand(List<DriverCommand> driverCommandList){
+        List<DriverCommand> commands = new ArrayList<>();
+        for (DriverCommand command : driverCommandList) {
+            try {
+                commands.add(command.clone());
+            } catch (CloneNotSupportedException e) {
+                e.printStackTrace();
```
kowallus: not really a nice way to handle a lack of cloning. 
It would be better to add a command rather than ignore it or throw another exception (since there is a risk of mutability).

kowallus (changes requested):
You should handle exceptions carefully.

src/main/java/edu/kis/powp/jobs2d/command/ImmutableComplexCommand.java
michaaeu: Done, but is it really a good solution? If there is a problem with cloning, a potentially mutable element will be added to the list even though we wanted to avoid it.

kowallus: I agree. The propagation of an exception (e.g., ImmutableCommandCreationException) would be better.

kowallus:
There are still some conflicts, so I cannot perform merging.

PR #107, opened 2020-06-18
CommandFactory

Nyitz:


src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,83 @@
...
+            throw new IllegalArgumentException("Name not found");
+        }
+
+        if (resultCommand instanceof CompoundCommand) {
+            ((CompoundCommand) resultCommand).setName(commandName);
+        }
+        return resultCommand;
+    }
+
+    public void setCommand(DriverCommand command) {
```
kowallus: addCommand would be less confusing

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,83 @@
...
+        commandsMap.put(command.toString(), command);
```
kowallus: Command added to the factory should also be cloned, so it wouldn't be changed externally.
This would make possible to avoid CloneNotSupportedException in getCommand (since clones of cloneables should be cloneable).

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -63,68 +65,84 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
+		JList commandFactoryList = new JList(factoryModel);
+		JButton btnRunCommandFromFactory = new JButton("Run command");
+		JButton btnRemoveCommandFromFactory = new JButton("Remove command");
+		btnRunCommandFromFactory.addActionListener((ActionEvent event) -> {
+			if (commandFactoryList.getSelectedIndex() != -1) {
+				try {
+					CommandFactory.getInstance().runCommand((String) commandFactoryList.getSelectedValue());
+				} catch (CloneNotSupportedException e1) {
+					e1.printStackTrace();
+				}
```
kowallus: It is hard to understand how:
"CommandFactory.getInstance().runCommand((String) commandFactoryList.getSelectedValue())" could cause CloneNotSupportedException.
Exception should be avoided or changed to a different (maybe new) one.

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,83 @@
...
+        this.window = window;
+    }
+
+    @Override
+    public void update() {
+        DriverCommand command = driverCommandManager.getCurrentCommand();
+        String commandName = driverCommandManager.getCurrentCommandString().replace("CompoundCommand name: ", "");
+        commandsMap.put(commandName, command);
+        window.addCommandToFactoryList(commandName);
+    }
```
kowallus: I would prefer that such behavior should be optional. By default user should decide whether he/she wants to add a current command to the factory.

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -63,68 +65,84 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
 		CommandHistoryController.setListModel(model);
 		JList commandHistoryList = new JList(model);
 		commandHistoryList.setSelectionMode(ListSelectionModel.SINGLE_INTERVAL_SELECTION);
-		commandHistoryList.addListSelectionListener(new SelectHistoryListOptionListener(commandManager, commandHistoryList));
+		commandHistoryList
+				.addListSelectionListener(new SelectHistoryListOptionListener(commandManager, commandHistoryList));
+		c.weighty = 6;
+		content.add(new JScrollPane(commandHistoryList), c);
 
+		factoryModel = new DefaultListModel();
```
kowallus, line 112: this method is large. Please, split it into several consistent methods?

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -63,68 +65,84 @@ public CommandManagerWindow(DriverCommandManager commandManager) {
...
+		JList commandFactoryList = new JList(factoryModel);
+		JButton btnRunCommandFromFactory = new JButton("Run command");
```
kowallus: I would limit to setting the command (or both options: "set" and "run")

src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -135,6 +136,9 @@ public void run() {
 				CommandsFeature.setupCommandManager();
 				MacroFeature.setupMacroDriver();
 
+				CommandFactory.getInstance();
```
kowallus: Illogical line. Change the design so it is easy to understand (clean code is self expaining)

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Map;
+import java.util.Set;
+import edu.kis.powp.jobs2d.command.CompoundCommand;
+import edu.kis.powp.jobs2d.command.DriverCommand;
+import edu.kis.powp.jobs2d.command.gui.CommandManagerWindow;
+import edu.kis.powp.jobs2d.command.manager.DriverCommandManager;
+import edu.kis.powp.observer.Subscriber;
+
+public class CommandFactory implements Subscriber {
+    public Map<String, DriverCommand> commandsMap = new HashMap<>();
```
kowallus: It should be private

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,83 @@
...
+        DriverCommand resultCommand=null;
+        if(commandsMap.keySet().contains(commandName)){
+            resultCommand = commandsMap.get(commandName).clone();
+        }else{
+            throw new IllegalArgumentException("Name not found");
+        }
+
+        if (resultCommand instanceof CompoundCommand) {
+            ((CompoundCommand) resultCommand).setName(commandName);
+        }
```
kowallus: instanceof often points to a flaw in OO design. The cloning should be responsible for making a copy of a command name. Fix the problem, not the symptom.

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,83 @@
...
+import edu.kis.powp.jobs2d.command.CompoundCommand;
+import edu.kis.powp.jobs2d.command.DriverCommand;
+import edu.kis.powp.jobs2d.command.gui.CommandManagerWindow;
+import edu.kis.powp.jobs2d.command.manager.DriverCommandManager;
+import edu.kis.powp.observer.Subscriber;
+
+public class CommandFactory implements Subscriber {
+    public Map<String, DriverCommand> commandsMap = new HashMap<>();
+    private DriverCommandManager driverCommandManager;
+    private CommandManagerWindow window;
```
kowallus: There are too many dependencies and responsibilities in this class. Try to modify the code to extract dependencies to GUI.

kowallus (changes requested):
A good starting point, but there are many things to clean up. Later, you can integrate the factory with import/export feature.

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,51 @@
+package edu.kis.powp.jobs2d.features;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import edu.kis.powp.jobs2d.command.DriverCommand;
+import edu.kis.powp.jobs2d.command.gui.CommandManagerWindow;
```
kowallus, line 7: dependencies are removed, forgotten import

kowallus (approved):
Very simple factory, but the design is now better.

PR #108, opened 2020-06-19
InkUsageMonitor

Blackwatch98:
We have made changes requred in last PR. Waiting for next review.

Group:
Daniel Duczymiński 215710
Mateusz Domalążek 215705


src/test/java/edu/kis/powp/jobs2d/TestJobs2dApp.java
```diff
@@ -66,10 +67,12 @@ private static void setupDrivers(Application application) {
 		driver = new LineDriverAdapter(drawerController, LineFactory.getSpecialLine(), "special");
 		DriverFeature.addDriver("Special line Simulator", driver);
 		DriverFeature.updateDriverInfo();
+
+        InkUsageMethodObserver obs = new InkUsageMethodObserver(10000f);
```
kowallus: spaces instead of tabs (again!)

src/main/java/edu/kis/powp/jobs2d/command/gui/RefuelInkWindow.java
```diff
@@ -0,0 +1,60 @@
...
+
+import edu.kis.powp.appbase.gui.WindowComponent;
+import edu.kis.powp.jobs2d.drivers.InkUsageDriverDecorator;
+import javax.swing.*;
+import java.awt.*;
+import java.awt.event.ActionEvent;
+
+public class RefuelInkWindow extends JFrame implements WindowComponent
+{
+    private InkUsageDriverDecorator driver;
```
kowallus: Lack of ISP. RefuelInkWindow only uses restoreInk, but this method is tangled with Jobs2dDriver API.

src/main/java/edu/kis/powp/jobs2d/drivers/InkUsageDriverDecorator.java
```diff
@@ -0,0 +1,110 @@
...
+    public void setPosition(int x, int y) {
+        double wasted = inkCounter(this.x0, x, this.y0, y);
+        if((inkLimit-wasted) >= 0){
+            driver.setPosition(x, y);
+            this.x0 = x;
+            this.y0 = y;
+        }
+    }
+
+    public void restoreInk(double amount) {
```
kowallus: should be declared in a separate interface

src/main/java/edu/kis/powp/jobs2d/drivers/InkUsageMethodObserver.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public InkUsageMethodObserver(double inkLimit){
+        previousIUD = new InkUsageDriverDecorator(DriverFeature.getDriverManager().getCurrentDriver(), inkLimit);
+    }
+
+    @Override public void update()
+    {
+        Job2dDriver currentDriver = DriverFeature.getDriverManager().getCurrentDriver();
+        InkUsageDriverDecorator driver = new InkUsageDriverDecorator(currentDriver, previousIUD.getInkLimit(),
+                previousIUD.getTotalUsage(), previousIUD.getMaxInkLimit());
```
kowallus: wouldn't it be easier to have setDriver in InkUsageDriverDecorator??

src/main/java/edu/kis/powp/jobs2d/drivers/InkUsageDriverDecorator.java
```diff
@@ -0,0 +1,110 @@
...
+        this.maxInkLimit = maxInkLimit;
+        pub = new Publisher();
+        pub.addSubscriber(new NoInkObserver(this));
+    }
+
+
+    @Override
+    public void setPosition(int x, int y) {
+        double wasted = inkCounter(this.x0, x, this.y0, y);
+        if((inkLimit-wasted) >= 0){
```
kowallus: setPosition does not affect INK (operation should be executed unconditionally), but it should affect the totalUsage.

kowallus (changes requested):
Better, but there are some logical and clean code issues (some are only suggestions).

kowallus:
It is nice that you added InkOperator interface but it should be used somewhere instead of 
 InkUsageDriverDecorator.

src/main/java/edu/kis/powp/jobs2d/drivers/InkUsageDriverDecorator.java
Blackwatch98: When it comes to this point we assumed that totalUsage is measure for whole Ink used by this machine (not the head) but anyway we adjusted it to Yours suggestion.

Blackwatch98:
The PR is once more ready for the review

src/main/java/edu/kis/powp/jobs2d/drivers/InkUsageDriverDecorator.java
kowallus: I agree with you: I didn't want to add: totalUsage += wasted.
 I meant that 'if' statement seemed to be unnecessary.

src/main/java/edu/kis/powp/jobs2d/drivers/InkUsageDriverDecorator.java
```diff
@@ -0,0 +1,102 @@
...
+    {
+        this.driver = driver;
+        pub = new Publisher();
+        pub.addSubscriber(new NoInkObserver(this));
+    }
+
+    @Override
+    public void setPosition(int x, int y) {
+        double wasted = inkCounter(this.x0, x, this.y0, y);
+        totalUsage += wasted;
```
kowallus: If totalUsage concerns only "ink", this line should be deleted.

src/main/java/edu/kis/powp/jobs2d/command/gui/RefuelInkWindow.java
kowallus: You did not understand my suggestion (you have made many unnecessary changes). ISP (Interface Segregation Principle) issue concerned the driver field. Earlier driver was used only for restoreInk, now it is used only for setInkLimit. What I wanted you to do was to change this line to:
private InkOperator operator
I wanted to add InkOperator interface to InkUsageDriverDecorator so it could be used here without any other changes to RefuelInkWindow or InkUsageDriverDecorator interface.

kowallus (commented):
You misinterpreted some of my comments. The changes to your code did not fix ISP. Most of the modifications wasn't necessary. To fix ISP you should only change the declaration of the driver field in RefuelInkWindow. If you will revert unwanted changes and fix it I can give you 4 for project.

Blackwatch98:
Ok tomorrow we will try to fix this

kowallus:
ok

PR #110, opened 2020-06-22
[IO-1] Transformation on a driver level

spacja255:
Hubert Bast 215674
Przemysław Frąg 215719


pom.xml
```diff
@@ -14,8 +14,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.0</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>10</source>
+                    <target>10</target>
```
kowallus: please do not change java version to higher

src/main/java/edu/kis/powp/jobs2d/command/ImmutableComplexCommand.java
```diff
@@ -17,7 +20,13 @@
      * @param commands - List containing DriverCommands
      */
     public ImmutableComplexCommand(List<DriverCommand> commands) {
-        driverCommands = Collections.unmodifiableList(new ArrayList<DriverCommand>(commands));
+        ArrayList<DriverCommand> list = new ArrayList<>(commands.size());
```
kowallus: Declaration should focus on the interface, i.e. List

src/main/java/edu/kis/powp/jobs2d/drivers/transformation/TransformationDriver.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.ArrayList;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+/**
+ * Transformation driver
+ */
+public class TransformationDriver implements Job2dDriver {
+    private final Job2dDriver driver;
+    private final ArrayList<Transformation> allTransformations = new ArrayList<>();
```
kowallus: declaration: List

src/main/java/edu/kis/powp/jobs2d/drivers/transformation/TransformationDriver.java
```diff
@@ -0,0 +1,56 @@
+package edu.kis.powp.jobs2d.drivers.transformation;
+
+import java.util.ArrayList;
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+/**
+ * Transformation driver
+ */
+public class TransformationDriver implements Job2dDriver {
```
kowallus, line 14: You should add the name of the pattern that is used in this class (to explain how to use it)

kowallus (changes requested):
The idea seems to be ok. It is tested partially, however you should implement it in the GUI application. 
There are also minor issues, making merging impossible.

PR #111, opened 2020-06-22
[IO-3] Complex command editor && Extending CM window 'reset observers' button fix

Lisek8:
Jędrzej Lis-Marciniak 215799
David Zientak 215934

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
@@ -15,14 +15,25 @@ public interface ICompoundCommand extends DriverCommand, Iterable<DriverCommand>
 	@Override
 	default DriverCommand clone() throws CloneNotSupportedException{
 		List<DriverCommand> commands = new ArrayList<>();
-
 		for (DriverCommand c : this) {
 			commands.add(c.clone());
 		}
 
 		return new CompoundCommand(commands);
   }
-    
+
+	default ICompoundCommand moveUpCommand(DriverCommand command){
+		return this;
+	}
+
+	default ICompoundCommand moveDownCommand(DriverCommand command){
+		return this;
+	}
+
+	default ICompoundCommand changeCoordinates(DriverCommand commandToChange, DriverCommand newCommand){
+		return this;
+	}
+
 	default public void accept(CommandVisitorInterface visitor) {
 		visitor.visit(this);
 	}
--- a/src/main/java/edu/kis/powp/jobs2d/command/gui/CommandEditorWindow.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/gui/CommandEditorWindow.java
@@ -1,0 +1,150 @@
+package edu.kis.powp.jobs2d.command.gui;
+
+import edu.kis.powp.appbase.gui.WindowComponent;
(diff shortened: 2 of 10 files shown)


src/main/java/edu/kis/powp/jobs2d/command/gui/CommandEditorWindow.java
```diff
@@ -0,0 +1,150 @@
...
+        c.gridy = 5;
+        c.fill = GridBagConstraints.BOTH;
+        changeCoordinates = new JButton("Change coordinates ");
+        changeCoordinates.setPreferredSize(new Dimension(300, 30));
+        changeCoordinates.addActionListener((ActionEvent e) -> this.changeCoordinates(Integer.parseInt(coordX.getText()), Integer.parseInt(coordY.getText()),commandPartsList.getSelectedValue()));
+        content.add(changeCoordinates, c);
+    }
+
+    public void changeCoordinates(int x, int y, DriverCommand command) {
+        if(commandManager.getCurrentCommand() instanceof ICompoundCommand) {
```
kowallus, line 94: instanceof is a code-smell of a bad OO design. You should think of the way, to solve this issue (maybe the visitor pattern?)

src/main/java/edu/kis/powp/jobs2d/command/ICompoundCommand.java
```diff
@@ -15,14 +15,25 @@
...
+	}
+
+	default ICompoundCommand moveDownCommand(DriverCommand command){
+		return this;
+	}
+
+	default ICompoundCommand changeCoordinates(DriverCommand commandToChange, DriverCommand newCommand){
+		return this;
+	}
+
```
kowallus, line 36: You should not pollute the main interface with methods that introduce mutability. This solution must be revised.

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandEditorWindow.java
```diff
@@ -0,0 +1,150 @@
...
+    private void addCommandsToList(DefaultListModel<DriverCommand> listModel, DriverCommand command) {
+        if (command instanceof ICompoundCommand) {
+            ((ICompoundCommand) command).iterator().forEachRemaining(cmd -> {
+                if (cmd instanceof ICompoundCommand) {
+                    addCommandsToList(listModel, cmd);
+                } else {
+                    listModel.addElement(cmd);
+                }
+            });
+        }
```
kowallus, line 148: The same issue as above. Maybe instanceof can be avoided by making sure of that the method is working on mutable commands instances (not DriverCommand in general)?

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandEditorWindow.java
```diff
@@ -0,0 +1,150 @@
...
+        updateCurrentCommandField();
+    }
+
+    @Override
+    public void HideIfVisibleAndShowIfHidden() {
+        if (this.isVisible()) {
+            this.setVisible(false);
+        } else {
+            this.setVisible(true);
+        }
```
kowallus, line 125: this.setVisible(!this.isVisible())
:)

kowallus (changes requested):
The solution makes too many changes in already established components.

PR #112, opened 2020-06-22
IO-1 command editor

bopke:


src/main/java/edu/kis/powp/jobs2d/command/CommandEditor.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public void changeCoordinates(int x, int y) throws ImproperCommandTypeException {
+        if (command instanceof ICompoundCommand) {
+            throw new ImproperCommandTypeException();
+        }
+        if (command instanceof SetPositionCommand) {
+            command = new SetPositionCommand(x, y);
+        } else if (command instanceof OperateToCommand) {
+            command = new OperateToCommand(x, y);
+        }
```
kowallus: instanceof is a code smell of a bad OO design. Try to eliminate it. Maybe visitor can be used?

kowallus (changes requested):
The solution should be added to GUI, to enable manipulation of any command (available using driverCommandManager.getCurrentCommand)

kowallus:
I gave you 3 for the project. If you finish GUI I will consider 3.5

PR #114, opened 2020-06-23
Io 1 dev (history of used commands)

arekdziuda:
Arkadiusz Dziuda 215714
Alan Franas 215717


src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryWindowCommandChangeObserver.java
```diff
@@ -0,0 +1,17 @@
+package edu.kis.powp.jobs2d.command.historyComponent;
+
+import edu.kis.powp.observer.Subscriber;
+
+public class HistoryWindowCommandChangeObserver implements Subscriber {
+
+    private HistoryManagerWindow backLogManagerWindow;
```
kowallus, line 7: Two things: 
1. This is a remain of plagiarism (#63).
2. You haven't managed to implement a suggestion:
"Try to redesign the history component to be independent of the GUI. (e.g. we might want to change GUI or use history from another component without accessing GUI)." 
Now, this is obligatory. There should be another way to add or take something from history than accessing GUI.

kowallus (commented):
A refactor request.

kowallus:
Despite the merge, there is an obligatory refactor request.

src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryWindowCommandChangeObserver.java
arekdziuda: We haven't changed this file since ours previous PR. (https://github.com/iis-powp-2020/powp_jobs2d_projects/pull/68)
I'm really sorry, but then it wasn't a plagiarism, but is it now?

PR #115, opened 2020-06-24
[IO-2] Complex command editor

Ard2148810:


Ard2148810:
Adrian Ałaszewski
Adrian Kowalski
Tobiasz Kubiak
Łukasz Zalewski

src/main/java/edu/kis/powp/jobs2d/command/OperateToCommand.java
```diff
@@ -36,4 +36,14 @@ public int getPosX() {
...
+
+	@Override
+	public int getX() {
+		return posX;
+	}
+
+	@Override
+	public int getY() {
+		return posY;
+	}
```
kowallus: Redundant methods, you should reuse getPosX and getPosY. If it is necessary you can modify HasCoordinates interface.

src/main/java/edu/kis/powp/jobs2d/command/HasCoordinates.java
```diff
@@ -0,0 +1,7 @@
+package edu.kis.powp.jobs2d.command;
+
+public interface HasCoordinates {
```
kowallus, line 3: Strange name for an interface, but I understand the idea. Maybe you can find a more conventional alternative?

src/main/java/edu/kis/powp/jobs2d/command/gui/ComplexCommandEditorWindow.java
```diff
@@ -0,0 +1,249 @@
...
+		updateJList(complexCommandEditor.getEditedComplexCommand());
+		commandList.setSelectedIndex(index - 1);
+	}
+
+	private void handleListSelectionEvent(ListSelectionEvent e) {
+		if (!e.getValueIsAdjusting()) {
+			int index = commandList.getSelectedIndex();
+			if(index > -1) {
+				DriverCommand driverCommand = listModel.getElementAt(index);
+				if (driverCommand instanceof HasCoordinates) {
```
kowallus, line 209: using instanceof is a code small of a bad OO design. It would be nice to have a better solution (maybe using a visitor pattern or providing a specialized commands hierarchy for command editor purposes)

src/main/java/edu/kis/powp/jobs2d/command/gui/ComplexCommandEditorWindow.java
```diff
@@ -0,0 +1,249 @@
...
+	private final JList<DriverCommand> commandList;
+	private final DefaultListModel<DriverCommand> listModel;
+	private final JLabel numberOfCommandsValue;
+	private final JLabel commandLengthValue;
+
+	private IComplexCommandEditor complexCommandEditor;
+	private ICompoundCommand currentCommand;
+
+	public ComplexCommandEditorWindow() {
+
```
kowallus, line 31: This method is too large. It should be decomposed to several methods to make it easier to understand and handle.

kowallus (changes requested):
Good job. 
Still, there are some small issues preventing from merging.

PR #117, opened 2020-06-24
IO-1 History of used commands (redesigned)

arekdziuda:
We have implemented suggestion to make history component to be independent of the GUI

Arkadiusz Dziuda 215714
Alan Franas 215717

### DIFF
--- a/src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryManager.java
+++ b/src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryManager.java
@@ -1,0 +1,30 @@
+package edu.kis.powp.jobs2d.command.historyComponent;
+
+import edu.kis.powp.jobs2d.command.manager.DriverCommandManager;
+
+import javax.swing.*;
+import java.time.format.DateTimeFormatter;
+
+public class HistoryManager {
+
+    private static final String HISTORY_TIME_PATTERN = "HH:mm:ss";
+    public static DefaultListModel<CommandHistory> historyEntryDefaultListModel;
+    public static DriverCommandManager commandManager;
+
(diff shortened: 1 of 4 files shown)


src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryManager.java
```diff
@@ -0,0 +1,30 @@
...
+
+import edu.kis.powp.jobs2d.command.manager.DriverCommandManager;
+
+import javax.swing.*;
+import java.time.format.DateTimeFormatter;
+
+public class HistoryManager {
+
+    private static final String HISTORY_TIME_PATTERN = "HH:mm:ss";
+    public static DefaultListModel<CommandHistory> historyEntryDefaultListModel;
```
kowallus, line 11: should be private (injected or accessed through getter)

src/main/java/edu/kis/powp/jobs2d/command/historyComponent/HistoryManager.java
```diff
@@ -0,0 +1,30 @@
...
+    public static DriverCommandManager commandManager;
+
+    public HistoryManager(DriverCommandManager commandManager) {
+        historyEntryDefaultListModel = new DefaultListModel<>();
+        HistoryManager.commandManager = commandManager;
+    }
+
+    public void updateCommandListField() {
```
kowallus, line 19: if "commandManager.getCurrentCommandString(), commandManager.getCurrentCommand()" were passed in params. The history could be also independent of commandManager. Extracting dependency to the caller does not introduce any new dependencies.

kowallus (approved):
OK. Just a minor suggestion for lower coupling.

PR #118, opened 2020-06-26
Fixed suspicious Class issue, added new function

PrzemaXster:
Added SelectClearMacro class to event package, which causes MacroAdapter driver to forget previously saved command


PrzemaXster:
My mistake, I will try again

PR #119, opened 2020-06-29
Added Macro Functionality that saves and runs commands

PrzemaXster:
Created classes:
-MacroFeature
-MacroAdapter
-SelectClearMacroDriver
-SelectLoadMacroDriver


src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverDecorator.java
```diff
@@ -10,22 +10,18 @@
...
 
-    public MacroAdapter() {
+    public MacroDriverDecorator() {
         this.driverCommandList = new ArrayList<>();
     }
 
-    public MacroAdapter(List<DriverCommand> driverCommandList) {
-        this.driverCommandList = driverCommandList;
+    public void setDriver() {
+        this.driver = DriverFeature.getDriverManager().getCurrentDriver();
```
kowallus, line 22: Unnecessary dependency to DriverManager and DriverFeature. Driver should be injected (as a constructor param).

src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverDecorator.java
```diff
@@ -38,22 +34,21 @@ public void clearCommandSet() {
         return driverCommandList;
     }
 
-    public void setDriverFeatureCurrentDriver() {
-        DriverFeature.getDriverManager().setCurrentDriver(driver);
-    }
 
     @Override
     public void setPosition(int i, int i1) {
```
kowallus, line 39: names o params should be x, y

kowallus (approved):
OK (minor issues in comments)

PR #122, opened 2020-09-10
Io 1 dev Usable Command Factory

Kacpri:


kowallus (commented):
What are these methods for?

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -4,107 +4,150 @@
...
+        );
+        addToWindow(btnSetSelectedCommand);
+        btnRunSelectedCommand = new JButton("Run selected command");
+        btnRunSelectedCommand.addActionListener((ActionEvent e) ->
+                {
+                    commandManager.runCurrentCommand();
+                }
+        );
+        addToWindow(btnRunSelectedCommand);
     }
```
kowallus: This method is too large. Should be split into several methods to make it easier to comprehend and manage this factory feature.

src/main/java/edu/kis/powp/jobs2d/command/gui/CommandManagerWindow.java
```diff
@@ -175,6 +219,19 @@ private void addObservers(List<Subscriber> observers) {
...
+
+    private void deleteSelectedCommand() {
+    }
+
+    private void setSelectedCommand() {
+    }
+
+    private void runSelectedCommand() {
+    }
+
```
kowallus: What are these methods for?

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -0,0 +1,36 @@
+package edu.kis.powp.jobs2d.features;
+
+import edu.kis.powp.jobs2d.command.DriverCommand;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class CommandFactory {
```
kowallus: It would be nice to have some "openness" for serialization and deserialization.

kowallus (changes requested):
Please, first, deal with those comments.

PR #124, opened 2020-09-15
Fix macro recording

tmajcher:


src/main/java/edu/kis/powp/jobs2d/drivers/macro/MacroDriverObserver.java
```diff
@@ -0,0 +1,23 @@
...
+import java.util.List;
+
+public class MacroDriverObserver implements Subscriber {
+
+    @Override
+    public void update() {
+        Job2dDriver currentDriver = DriverFeature.getDriverManager().getCurrentDriver();
+        if (!(currentDriver instanceof MacroDriverDecorator) && MacroFeature.getMacroDriverDecorator().isFlagActive()) {
+            MacroDriverDecorator driver = MacroFeature.getMacroDriverDecorator();
+            driver.setDriver();
```
kowallus: This part of the code is hard to decode. Moreover, MacroDriverDecorator seems to be unnecessarily tangled with DriverManager and DriverFeature (see: DriverFeature.getDriverManager() call in setDriver implementation). 
Please clean up those parts of the code.

kowallus (changes requested):
Only one comment to fix

kowallus (approved):
OK (but you forget to remove import statement ;) )

PR #125, opened 2020-09-17
Saving and loading CommandFactory to/from file.

Nyitz:


src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -48,4 +56,28 @@ public void removeCommand(String commandName) {
...
+            for (String item : commandsMap.keySet()) {
+                bufferedWriter.write("{\"" + item + "\":" + (new JsonParser()).parseToString(commandsMap.get(item)) + "}" + "\n");
+            }
+		} catch (IOException exception) {
+			exception.printStackTrace();
+		}
+    }
+
+    private static void loadFromFile() {
+        try (BufferedReader bufferedReader = new BufferedReader(new FileReader("command_factory"))) {
```
kowallus, line 71: Not clean: "command_factory" should be at least a constant. In this case, it would be better to have the name of the file injected or passed through a parameter (it should be configured in TestJobs2dApp.java).

BTW: You could also use a builder pattern to encapsulate this feature.

src/main/java/edu/kis/powp/jobs2d/features/CommandFactory.java
```diff
@@ -48,4 +56,28 @@ public void removeCommand(String commandName) {
...
         return commandsMap.keySet();
     }
+
+    private void saveToFile() {
+		try (BufferedWriter bufferedWriter = new BufferedWriter(new FileWriter(new File("command_factory")))) {
+            for (String item : commandsMap.keySet()) {
+                bufferedWriter.write("{\"" + item + "\":" + (new JsonParser()).parseToString(commandsMap.get(item)) + "}" + "\n");
+            }
+		} catch (IOException exception) {
+			exception.printStackTrace();
```
kowallus, line 66: Unhandled exceptions (ugly)

kowallus (commented):
No so clean but I accept. Please read comments.

PR #126, opened 2021-01-16
Cleaning up

TheDamik34:


src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -51,7 +50,7 @@ public synchronized void clearCurrentCommand() {
     }
 
     public synchronized void runCurrentCommand() {
-        currentCommand.execute(driverManager.getCurrentDriver());
+        currentCommand.execute(driverManager.getDriverToDraw());
```
kowallus: changing default API is troublesome for others. It would be better if getCurrentDriver() could stay here.

src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -1,34 +1,50 @@
...
 
 /**
  * Driver manager provides means to setup the driver. It also enables other
  * components and features of the application to react on configuration changes.
  */
 public class DriverManager {
-
-    private Job2dDriver currentDriver = new LoggerDriver();
+    private Job2dDriver currentDriver;
+    private ExtensionDriverComposite extensionDriver = new ExtensionDriverComposite();
```
kowallus: You should embed the current driver in the composite, so it could be managed there.

src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -1,34 +1,50 @@
...
      * @param driver Set the driver as current.
      */
     public synchronized void setCurrentDriver(Job2dDriver driver) {
+        Job2dDriver temp = currentDriver;
         currentDriver = driver;
-        changePublisher.notifyObservers();
+        extensionDriver.setDriver(currentDriver);
+        if(temp instanceof UsageMonitorExtension) {
+            changePublisher.notifyObservers();
+        }
```
kowallus: observers should work without such hacks. If really necessary at least you should introduce "instanceof" on some general interface.

src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -1,34 +1,50 @@
...
     /**
      * @return Current driver.
      */
+    public synchronized Job2dDriver getDriverToDraw() {
+        if(extensionDriver.getExtensions().size() == 0) {
+            return currentDriver;
+        }
+        return extensionDriver;
+    }
+
```
kowallus: This code should be inside the composite class.

src/main/java/edu/kis/powp/jobs2d/drivers/UsageMonitorObserver.java
```diff
@@ -1,15 +1,16 @@
...
 
     @Override public void update() {
         Job2dDriver currentDriver = DriverFeature.getDriverManager().getCurrentDriver();
-        if (!(currentDriver instanceof Job2dDriverDecorator)) {
-            DriverFeature.getDriverManager().setCurrentDriver(new Job2dDriverDecorator(currentDriver));
+        if (!(currentDriver instanceof UsageMonitorExtension)) {
+            DriverFeature.getDriverManager().setCurrentDriver(new UsageMonitorExtension());
         }
     }
 }
```
kowallus: I think that this code should be replaced by some management features inside your composite.

src/main/java/edu/kis/powp/jobs2d/drivers/extensions/ExtensionDriverComposite.java
```diff
@@ -0,0 +1,46 @@
+package edu.kis.powp.jobs2d.drivers.extensions;
+
+import edu.kis.powp.jobs2d.Job2dDriver;
+
+import java.util.*;
+
+public class ExtensionDriverComposite implements ExtensionDriver {
+    private Map<String, ExtensionDriver> extensions = new HashMap<>();
+    private Job2dDriver driver;
```
kowallus, line 9: what about a list of "horizontal" extensions (not decorators).

kowallus (changes requested):
I think you missed the goal. Check the suggestions.

src/main/java/edu/kis/powp/jobs2d/drivers/UsageMonitorObserver.java
TheDamik34: I think it should be how it is, its part of an observer. not of my composite implementation. I have just changed the name of Job2dDriverDecorator to UsageMonitorExtension and made it fluent to use any base driver.

src/main/java/edu/kis/powp/jobs2d/drivers/extensions/ExtensionDriverComposite.java
TheDamik34: I guess that i didnt get what you have meant by "horizontal" extensions. I have a map of extensions which are UsageMonitorExtension and  TransformationDriver (should be called TransformationExtension). The driver describes a type of line to use by extensions. I dont see those "horizontal" extensions, could you explain more?

src/main/java/edu/kis/powp/jobs2d/command/manager/DriverCommandManager.java
```diff
@@ -50,7 +51,7 @@ public synchronized void clearCurrentCommand() {
     }
 
     public synchronized void runCurrentCommand() {
-        currentCommand.execute(driverManager.getDriverToDraw());
+        currentCommand.execute(driverManager.getExtensionDriver().getCurrentDriver());
```
kowallus: You should not change the API. Originally there was getCurrentDriver() and it should be enough:

- getCurrentDriver should return driver composite and it should simply be able to delegate all requests to child (horizontal or extension) sub-drivers.
- setCurrentDriver should delegate a request to set driver to driver composite so it could manage it. 
- you might need to replece some getCurrentDriver and setCurrentDriver when they are used e.g. to apply decorator (because driver composite with your extensions should be able to manage them)

src/main/java/edu/kis/powp/jobs2d/drivers/DriverManager.java
```diff
@@ -10,34 +10,15 @@
...
-        if(extensionDriver.getExtensions().size() == 0) {
-            return currentDriver;
-        }
-        return extensionDriver;
-    }
-
-    public synchronized Job2dDriver getCurrentDriver() {
-        return currentDriver;
+        extensionDriver.setDriver(driver);
+        changePublisher.notifyObservers();
```
kowallus: This looks much better

kowallus (changes requested):
It looks better now, but I still want one fix, and then we can talk.

kowallus:
OK, it is enough. Now, horizontal extensions seem to work correctly and the design is acceptable.

